PR #2, opened 2020-04-03
Allow multiple pools to be started for each {scheme, host, port}

sneako:
Here is my initial implementation of starting multiple pools.

Main changes:

- PoolRegistry changed to `keys: :duplicate`
- Pool size & pool count are configured via application env. This is temporary while we figure out how we want to configure the lib. Ideally, we would be able to provide a default configuration, and override that config for specific `{s, h, p}`s
- Conn module will never return a bare Mint struct, always a Finch.Conn

I added one really basic test for now, just so that I had a quick sanity check while I was moving things around.

There is currently a race condition, which I noticed while benchmarking. If more than one request comes in before a pool finishes registering itself, multiple group of pools will be started. I got around that in the benchmarking by just making a single request (to create the pools) before starting the benchmark.

I have been exercising it locally using this:
https://github.com/sneako/httpc_bench/tree/finch

I am planning to write a terraform plan that will allow us to easily spin up a client & server in AWS for more realistic benchmarking.

Are you against running the formatter on the codebase?


lib/finch/pool.ex
```diff
@@ -25,11 +23,19 @@ defmodule Finch.Pool do
     receive_timeout = Keyword.get(opts, :receive_timeout, 15000)
 
     NimblePool.checkout!(pool, :checkout, fn {conn, pool} ->
+      # do we need this try catch?
```
josevalim, line 26: It depends. It may not be necessary as long as we can close the connection from a process that does not own it. You can test it by doing something like this in IEx:

```
conn = Conn.connect(conn)
spawn fn -> Conn.close(conn) end
```

If the connection is closed afterwards, then you can remove the try/catch and the pool will close it for you. If you need to close it in the caller though, then you need to keep it. Please keep me posted as I should update NimblePool's readme too.

sneako: It appears that you do not need to be the owner to close a Conn:
```
iex(1)> alias Finch.Conn                                      
Finch.Conn
iex(2)> conn = Conn.new(:https, "github.com", 443, [], self())
%{
  host: "github.com",
  mint: nil,
  opts: [],
  parent: #PID<0.311.0>,
  port: 443,
  scheme: :https
}
iex(3)> conn = Conn.connect(conn)                             
%{
  host: "github.com",
  mint: %Mint.HTTP1{
    buffer: "",
    host: "github.com",
    mode: :active,
    port: 443,
    private: %{},
    request: nil,
    requests: {[], []},
    scheme_as_string: "https",
    socket: {:sslsocket, {:gen_tcp, #Port<0.11>, :tls_connection, :undefined},
     [#PID<0.319.0>, #PID<0.318.0>]},
    state: :open,
    transport: Mint.Core.Transport.SSL
  },
  opts: [],
  parent: #PID<0.311.0>,
  port: 443,
  scheme: :https
}
iex(4)> spawn fn -> Conn.close(conn) |> IO.inspect end        
#PID<0.321.0>
%{      
  host: "github.com",
  mint: %Mint.HTTP1{
    buffer: "",
    host: "github.com",
    mode: :active,
    port: 443,
    private: %{},
    request: nil,
    requests: {[], []},
    scheme_as_string: "https",
    socket: {:sslsocket, {:gen_tcp, #Port<0.11>, :tls_connection, :undefined},
     [#PID<0.319.0>, #PID<0.318.0>]},
    state: :closed,
    transport: Mint.Core.Transport.SSL
  },
  opts: [],
  parent: #PID<0.311.0>,
  port: 443,
  scheme: :https
}
```

josevalim: So I think we are good to go without it!

sneako:
Just pushed a pretty big change, to allow user to configure the pools. Writing some tests for that now :)

lib/finch/pool_manager.ex
```diff
@@ -26,26 +23,43 @@ defmodule Finch.PoolManager do
...
+  def lookup_pool(registry, key) do
+    case Registry.lookup(registry, key) do
       [] ->
         :none
 
       [{pid, _}] ->
         {:ok, pid}
+
+      pids ->
+        {pid, _} = Enum.random(pids)
```
keathley, line 35: This is good for now. In the future, I think it would be nice to be able to plug in alternative strategies here.  But it's not something we need to worry about now.

lib/finch.ex
```diff
@@ -1,8 +1,48 @@
...
+      @config %{
+        name: __MODULE__,
+        pool_sup_name: :"#{__MODULE__}.PoolSup",
+        pool_reg_name: :"#{__MODULE__}.PoolRegistry",
+        pools: Keyword.get(unquote(opts), :pools, %{})
+      }
+
+      def start_link do
+        name = :"#{__MODULE__}.PoolManager"
+        Finch.PoolManager.start_link(name, @config)
```
keathley, line 37: What do you think about allowing the pool configuration to be passed in as part of `start_link`. That way users can set the pool sizes at runtime. I know that we would need to do this occasionally because we don't always know the location of a service at compile time. We have to get it from the environment after the app boots.

sneako: That makes sense. How about we allow both, and I will merge the runtime config with the compile time config here?

keathley: That seems perfect

keathley:
> Are you against running the formatter on the codebase?

I haven't been. I typically disable it by removing the files from the `.formatter.exs` because I don't run it locally. But, if people feel strongly we can enable it :+1:.

lib/finch.ex
```diff
@@ -1,8 +1,48 @@
 defmodule Finch do
   @moduledoc """
   Finch
+
+  Create your Finch HTTP Client like so:
+
+      defmodule MyFinch do
+        use Finch
```
josevalim, line 8: I would prefer if we did not tie Finch to modules. Instead, we could make it so `finch` always requires the name on start and APIs. Then we can start it like this in your sup tree:

```elixir
{Finch, name: MyFinch, pools: %{
   :default => %{size: 10, count: 1},
   {:https, "hex.pm", 443} => %{size: 20, count: 4}
}
```

And then `Finch.request(MyFinch, ...)`. If users want to wrap it though, it would be easy. Thoughts?

sneako: I can definitely see the appear of not being tied to a module, but I went this way because I wanted to avoid calculating the supervisor and registry names on every single interaction. Maybe that cost is not significant though.

I think we would also have to pass the pool config in for every request too, so that the PoolManager has access to it when it is starting the pools.

Realizing now that this means I cannot merge the runtime and compile time config in Finch.start_link and expect it to be used when calling Finch.request.

sneako: I guess passing in the pool config to every request wouldn't be that bad. And if a user really has a huge amount of {s, h, p} to configure, they could stick the pool config in a persistent_term, which we could mention in the docs

keathley: > I think we would also have to pass the pool config in for every request too, so that the PoolManager has access to it when it is starting the pools.

An alternative would be to stash the configuration in an ets table or persistent term and look it up using the `:name`.

sneako: ha, great minds!

sneako: Ok, i'm going to remove the __using__ macro and stick the configuration in a persistent term in start_link

josevalim: > I can definitely see the appear of not being tied to a module, but I went this way because I wanted to avoid calculating the supervisor and registry names on every single interaction. 

This is definitely a good point. What I usually do in this case is to start a ETS table with the same name and store the values in there. Since you also have a Registry, note the registry supports metadata, so you can name the registry with the given `name` and store this information in the registry metadata. This way you don't need to manage a separate ets table nor the persistent term storage.

sneako:
I have pushed the change from `__using__` to passing everything in to start_link, and storing the config in the Registry meta

sneako:
In order to get around the race condition I mentioned above, I am now starting the pools for each configured `{s, h, p}`, in Finch.start_link/1. However, the race condition is still possible with any non configured hosts.

josevalim:
Since pools are unlikely to be started dynamically multiple times, it may be worth having a genserver inside Finch with the purpose of serializing the pool initialization?

sneako:
Oh right, of course! I'll give that a try

sneako:
Now, the supervisor is in the Finch module, and the PoolManager has been changed to a GenServer, so that we can serialize requests to start pools through there.

I also switched from HTTParrot to Bypass and started adding a few more basic tests. It appears that HTTParrot doesn't even let you start multiple servers, so that is pretty much a deal breaker for us.

lib/finch.ex
```diff
@@ -26,8 +58,33 @@ defmodule Finch do
...
+    config = %{
+      registry_name: name,
+      manager_name: manager_name(name),
+      supervisor_name: pool_supervisor_name(name),
+      pools: pools
+    }
+
+    Supervisor.start_link(__MODULE__, config, name: supervisor_name(name))
+  end
+
```
josevalim, line 74:
```suggestion

@impl true
```

lib/finch/pool_manager.ex
```diff
@@ -26,26 +34,45 @@ defmodule Finch.PoolManager do
...
+      pids ->
+        # TODO implement alternative strategies
+        {pid, _} = Enum.random(pids)
+        {:ok, pid}
     end
   end
 
-  def start_pool(key) do
-    DynamicSupervisor.start_child(PoolSup, {Finch.Pool, key})
+  def start_pools(registry_name, shp) do
```
josevalim, line 52: Is this called from anywhere? 🤔

lib/finch.ex
```diff
@@ -26,8 +58,33 @@ defmodule Finch do
...
+  def init(config) do
+    children = [
+      {DynamicSupervisor, name: config.supervisor_name, strategy: :one_for_one},
+      {Registry, [keys: :duplicate, name: config.registry_name, meta: [config: config]]},
+      {PoolManager, config}
+    ]
+
+    Supervisor.init(children, strategy: :one_for_one)
```
josevalim, line 82: It should probably be `:one_for_all` (safer) until we can stop and think about the different failure modes and restarts.

lib/finch/pool_manager.ex
josevalim: Ignore. :)

lib/finch/pool_manager.ex
```diff
@@ -26,26 +34,45 @@ defmodule Finch.PoolManager do
...
+    Enum.map(1..count, fn _ ->
+      DynamicSupervisor.start_child(config.supervisor_name, {Finch.Pool, pool_args})
+    end)
+    |> hd()
+  end
+
+  defp pool_config(%{pools: config}, shp) do
+    case Map.get(config, shp, config[:default]) do
+      nil -> {1, 10}
+      %{count: count, size: size} -> {count, size}
```
josevalim, line 75: Note this requires the pool configuration to always have both keys. Maybe that's fine for now, just thought I would mention.

lib/finch/pool_manager.ex
```diff
@@ -26,26 +34,45 @@ defmodule Finch.PoolManager do
...
+    {:reply, reply, state}
+  end
+
+  defp do_start_pools(shp, config) do
+    {count, size} = pool_config(config, shp)
+    pool_args = {shp, config.registry_name, size}
 
-    Supervisor.init(children, strategy: :one_for_one)
+    Enum.map(1..count, fn _ ->
+      DynamicSupervisor.start_child(config.supervisor_name, {Finch.Pool, pool_args})
```
josevalim, line 67:
```suggestion
      {:ok, pid} = DynamicSupervisor.start_child(config.supervisor_name, {Finch.Pool, pool_args})
      pid
```

So we make sure they have all started properly?

lib/finch/pool_manager.ex
```diff
@@ -1,19 +1,27 @@
...
 
-  def get_pool(scheme, host, port) do
+  def get_pool(registry_name, scheme, host, port) do
     key = {scheme, host, port}
-    case lookup_pool(key) do
+
+    case lookup_pool(registry_name, key) do
       :none ->
-        case start_pool(key) do
+        case start_pools(registry_name, key) do
```
josevalim, line 24: `start_pools` should no longer return `{:error, _}`, correct? Because we have removed the race. So maybe we can remove the case?

lib/finch/pool_manager.ex
sneako: Good point, I'll make it start one pool by default, and then `:count` can be optional

lib/finch.ex
```diff
@@ -1,10 +1,43 @@
...
+  interact with.
+
+  You can also configure a pool size and count to be used for each specific
+  `{scheme, host, port}`s that are known before starting Finch. When configuring
+  pools, the `:size` is required. The `:count` will default to 1.
+
+  ```
+  Finch.start_link(name: MyConfiguredFinch, pools: %{
+    :default => %{size: 10},
+    {:https, "hex.pm", 443} => %{size: 32, count: 8}
```
sneako, line 30: Should we allow users to pass url binaries here too?

keathley:
This is great @sneako! I'll merge this in so we can start building off of this work.

sneako:
Thanks @keathley! I've been working on benchmarking on ec2 this evening, I'll have some results to share very soon!

PR #3, opened 2020-04-06
If the connection has been closed, don't try to transfer it back to the pool

sneako:


lib/finch/pool.ex
```diff
@@ -27,9 +27,13 @@ defmodule Finch.Pool do
         conn = Conn.connect(conn)
 
         with {:ok, conn, response} <- Conn.request(conn, req, receive_timeout),
+             {true, _} <- {Conn.open?(conn), response},
              {:ok, conn} <- Conn.transfer(conn, pool) do
           {{:ok, response}, conn}
         else
+          {false, response} ->
+            {{:ok, response}, conn}
```
sneako, line 35: hm actually, I should make sure the conn returned here is from line 29, currently it would be the one from line 27. Updating now

PR #10, opened 2020-04-07
Docs

josevalim:
* Use child_spec examples whenever possible
* Clarify that pools are not automatically terminated


sneako:
❤️ ❤️ ❤️

PR #11, opened 2020-04-08
Define Finch.Response and return it from Finch.request

### LINKED ISSUE #5: Return response as a struct?
I was wondering if we should add a response struct. The main benefit behind this struct is that it would allow users of the library to understand what keys are available on the response they get back.

QuinnWilton:
This resolves #5 

Thanks for making some issues on the repo! This issue seemed pretty trivial, so I thought I'd start here :)

I grabbed the typespecs from [Mint](https://github.com/elixir-mint/mint/blob/master/lib/mint/types.ex) to make sure they align with the types in `Finch.Response`.

I did not enforce any of the fields on the struct, because we receive the fields one at a time. If you'd like to enforce the fields on the struct, I'm happy to defer construction of the struct until we process the `:done` message.

In the case of `:data`, I defaulted to `""` because I think it makes for a cleaner API, but I understand if this is a controversial choice, and you'd prefer to be able to distinguish between an empty body, and no body.

Regardless, I'm happy to make any changes. I tried to stick to existing coding style where possible, but since the project is fairly small still, I wasn't sure where you stand on things like `alias __MODULE__`

### DIFF
--- a/lib/finch/conn.ex
+++ b/lib/finch/conn.ex
@@ -1,6 +1,7 @@
 defmodule Finch.Conn do
   @moduledoc false
 
+  alias Finch.Response
   alias Mint.HTTP
 
   def new(scheme, host, port, opts, parent) do
@@ -59,7 +60,7 @@ defmodule Finch.Conn do
   def request(conn, req, receive_timeout) do
     case HTTP.request(conn.mint, req.method, req.path, req.headers, req.body) do
       {:ok, mint, ref} ->
-        receive_response([], %{conn | mint: mint}, ref, %{}, receive_timeout)
+        receive_response([], %{conn | mint: mint}, ref, %Response{}, receive_timeout)
 
       {:error, mint, error} ->
         {:error, %{conn | mint: mint}, error}
@@ -85,12 +86,16 @@ defmodule Finch.Conn do
 
   defp receive_response([entry | entries], conn, ref, response, timeout) do
     case entry do
-      {kind, ^ref, value} when kind in [:status, :headers] ->
-        response = Map.put(response, kind, value)
+      {:status, ^ref, value} ->
+        response = %{response | status: value}
+        receive_response(entries, conn, ref, response, timeout)
+
+      {:headers, ^ref, value} ->
+        response = %{response | headers: value}
         receive_response(entries, conn, ref, response, timeout)
 
       {:data, ^ref, data} ->
-        response = Map.update(response, :data, data, &(&1 <> data))
+        response = append_data_chunk(response, data)
         receive_response(entries, conn, ref, response, timeout)
 
       {:done, ^ref} ->
@@ -99,5 +104,13 @@ defmodule Finch.Conn do
       {:error, ^ref, error} ->
         {:error, conn, error}
     end
+  end
+
+  defp append_data_chunk(%Response{body: nil} = response, data_chunk) do
+    %{response | body: data_chunk}
+  end
+
+  defp append_data_chunk(%Response{body: body} = response, data_chunk) when is_binary(body) do
+    %{response | body: body <> data_chunk}
   end
 end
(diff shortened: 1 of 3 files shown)


sneako:
Thanks! We are just getting started here, so it's exciting to already have people tackling the issues we just created yesterday :)

This looks great to me! However, I do think there is some value in being able to distinguish between no body at all, and an empty body, so I think we should default data to `nil`.

lib/finch/response.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Finch.Response do
+  alias __MODULE__
```
sneako, line 2: maybe since this module is not private, we should add a brief `@moduledoc`?

lib/finch/response.ex
```diff
@@ -0,0 +1,19 @@
...
+
+  defstruct [
+    :status,
+    headers: [],
+    data: ""
```
keathley, line 7: Should we go with `body` instead of `data`?

keathley:
This is looking great @QuinnWilton! I agree with @sneako that it's probably useful to return the body as `nil` for the reasons y'all stated. If there's a precedent against that I'm OK with being persuaded though.

lib/finch/response.ex
sneako: Good point. `data` makes sense within Mint, but at this level of abstraction `body` probably makes more sense

QuinnWilton:
All of those changes sound good! I'll get to them after I finish up my morning work routine. `body` is my preference too, but I didn't want to change things. It sounds like we're all in agreement there though!

QuinnWilton:
> If there's a precedent against that I'm OK with being persuaded though.

Less a precedent than being able to assume that the field is always present and containing a binary, versus needing to handle the `nil` case :) It's probably worth the extra overhead to give the caller more control over handling those sorts of weird cases though.

QuinnWilton:
I pushed up the changes :)

lib/finch/response.ex
sneako: Great job with these docs!

lib/finch/conn.ex
```diff
@@ -100,4 +105,12 @@ defmodule Finch.Conn do
         {:error, conn, error}
     end
   end
+
+  defp append_data_chunk(%Response{body: nil} = response, data_chunk) do
+    %{response | body: data_chunk}
+  end
+
+  defp append_data_chunk(%Response{body: body} = response, data_chunk) when is_binary(body) do
```
sneako, line 113: I don't feel too strongly about it, but this guard seems a bit superfluous to me. Maybe we should flip these clauses and remove the match on nil? What do you think?

QuinnWilton: I can do that! My intent was to detect misuse of the function, but re-reading this now I realize that I'd be better off having `when is_binary(data_chunk)` as the guard instead. Does flipping the two clauses, and adding that guard work for you?

keathley:
I should read through the spec probably. I _think_ I recall that the body is optional which is why I felt like `nil` might make sense here. I seem to recall that the Ruby HTTP client returns nil if the body is empty. That's not necessarily an indicator of anything. Just an anecdote. Maybe @whatyouhide or @ericmj have opinions on this. I'm sure they understand the HTTP spec way better than I do at this point ;)

lib/finch/response.ex
QuinnWilton: Thank you! They are heavily cribbed from the existing `Mint` docs

QuinnWilton:
> I should read through the spec probably. I _think_ I recall that the body is optional which is why I felt like `nil` might make sense here.

I think this is the relevant section of the spec: https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3

> The presence of a message-body in a request is signaled by the inclusion of a Content-Length or Transfer-Encoding header field in the request's message-headers.

In light of that, I think `nil` is the right choice here.

PR #15, opened 2020-04-08
Update CI checks

keathley:
This PR adds credo as a dependency and to our CI tests. I've used the credo file that I typically add to projects. Its less strict than a typical setup, but it enforces the stuff that I think typically make the most sense. I'm happy to discuss changes if y'all have a strong preference about any of this stuff.


sneako:
Great! I'm favor of this

keathley:
hmm...I'm not sure why that's failing on elixir 1.7.4. I'm guessing some sort of AST change or something? That `with` expression has 2 lines in it so it's not actually wrong.

sneako:
Hm why did credo pass on elixir 1.10, but not on 1.7 🤔

sneako:
Hm well I think the warning says it should have more than one <- clause, so it doesn't count the multiple else clauses. I wouldn't mind changing it to a three clause case

keathley:
I think its a bug. Or I'm misinterpreting the error. But that `with` clause has two expressions in it before the else.

.credo.exs
```diff
@@ -0,0 +1,158 @@
...
+      #
+      # If you create your own checks, you must specify the source files for
+      # them here, so they can be loaded by Credo before running the analysis.
+      #
+      requires: [],
+      #
+      # If you want to enforce a style guide and need a more traditional linting
+      # experience, you can change `strict` to `true` below:
+      #
+      strict: false,
```
antedeguemon, line 40:
```suggestion
      strict: true,
```

Maybe?

sneako:
Oh I was looking at master where it does only have one clause: https://github.com/keathley/finch/blob/master/lib/finch/pool.ex#L29

But here is this branch: https://sentry.io/organizations/appmonet/issues/1582780741/?project=1889179&query=is%3Aunresolved

sneako:
Looks like on "pull_request" it runs against the merged pr:
```
commit 6a67d3448636b5a7862d3d6b43479e06e4b38a24
Author: Chris Keathley <c@keathley.io>
Date:   Wed Apr 8 13:35:57 2020 +0000

    Merge 2e7be0c8c234640ade6f8cdb06294e728012f9a4 into 6bde388f8ca19e62420da5b7b6ac7e50749a6cfd
```
and on "push" it just checks out the current branch
```
commit 2e7be0c8c234640ade6f8cdb06294e728012f9a4
Author: Chris Keathley <c@keathley.io>
Date:   Wed Apr 8 09:35:30 2020 -0400

    Implement credo suggestions
```

keathley:
Ah duh. I didn't realize I hadn't rebased on master. Should be fixed up now.

PR #16, opened 2020-04-08
Remove unused error clause in Finch.Pool.request/3

QuinnWilton:
Finch.Conn.request/3 only returns error tuples of the form: `{:error, %Conn{}, term}`


QuinnWilton:
Yup, I think it was always a problem, but Dialyzer wasn't able to analyze the old `with` clause as easily as it can a `case` statement :)

PR #17, opened 2020-04-09
Fix race condition in PoolManager.start_pools/2

sneako:
I thought I had fixed this by serializing all pool starts through the GenServer, but I realized that the messages could still pile up in the PoolManager mailbox, and continue to be processed after the pools actually get started.


lib/finch/pool_manager.ex
```diff
@@ -51,7 +51,12 @@ defmodule Finch.PoolManager do
 
   @impl true
   def handle_call({:start_pools, shp}, _from, state) do
-    reply = do_start_pools(shp, state)
+    reply =
+      case lookup_pool(state.registry_name, shp) do
```
josevalim, line 55: Correct. We need to check them both on the server and client. Good catch!

PR #19, opened 2020-04-09
Reconnection Strategy

sneako:
Here is my initial implementation a reconnection strategy. I will add some tests shortly, but I wanted to open this up for feedback asap.

Currently if connection fails, it is retried immediately and perpetually until it succeeds, so this is should be a big improvement.

I used [backoff](https://github.com/ferd/backoff), so if connection fails we increment the backoff using `:backoff.rand_increment/1`, sleep and try again.

Most of the new code is around consuming & validating the configuration, so I think it could be really nice to switch to [NimbleOptions](https://github.com/dashbitco/nimble_options) once https://github.com/dashbitco/nimble_options/pull/21 is merged.

I changed the pool configuration to accept keyword lists instead of maps, I think that is more idiomatic than the maps I used initially, but I am open to switching back if we don't want to make that change. 

All pool configuration options are optional now, where previously `:size` was required.


lib/finch/conn.ex
```diff
@@ -9,22 +9,26 @@ defmodule Finch.Conn do
...
+
+  def connect(conn, prev_backoff) do
     with {:ok, mint} <- HTTP.connect(conn.scheme, conn.host, conn.port, conn.opts),
          {:ok, mint} <- HTTP.controlling_process(mint, conn.parent) do
       %{conn | mint: mint}
     else
       _ ->
-        conn
+        backoff = :backoff.rand_increment(prev_backoff, conn.backoff.max)
+        Process.sleep(backoff)
```
josevalim, line 30: Ideally you wouldn't sleep but rather send a message to yourself but we don't have a receive loop here, so the best option is to indeed to sleep for now. :)

sneako:
I've added some tests, but I think the easiest way to properly test the backoff will be to listen for telemetry events, which are currently being worked on.

test/finch_test.exs
```diff
@@ -20,21 +20,33 @@ defmodule FinchTest do
...
-      {:ok, %Response{}} = Finch.request(MyFinch, :get, endpoint(bypass), [], "")
+      {:ok, %Response{}} = Finch.request(MyFinch, :get, endpoint(bypass))
     end
 
     test "default can be configured", %{bypass: bypass} do
-      {:ok, _} = Finch.start_link(name: MyFinch, pools: %{default: %{count: 5, size: 5}})
+      {:ok, _} =
+        Finch.start_link(
+          name: MyFinch,
+          pools: %{default: [count: 5, size: 5, backoff: [size: 2, count: 4]]}
```
keathley, line 33: Is the `backoff` configuration here intentional? I'm not sure I'm following exactly what's happening in this test.

lib/finch.ex
```diff
@@ -131,6 +161,56 @@ defmodule Finch do
...
+    opts |> Keyword.get(:count, @default_pool_size) |> validate_pos_integer()
+  end
+
+  defp validate_backoff(nil) do
+    {:ok, %{initial: @default_backoff_initial, max: @default_backoff_max}}
+  end
+
+  defp validate_backoff(opts) do
+    init = Keyword.get(opts, :initial, @default_backoff_initial)
+    max = Keyword.get(opts, :initial, @default_backoff_max)
```
keathley, line 198: This should be `max` here right?

test/finch_test.exs
sneako: The values are arbitrary, and I added the configuration here, just to show that the default can be configured with all available options. But since I am not asserting that the values were actually set, I will remove them from here, and properly test backoff configuration after we merge the telemetry pr.

lib/finch.ex
sneako: ah of course

sneako: what do you think of `:initial`? Considering changing it to `:min`

test/finch_test.exs
keathley: Ah gotcha :) Thanks for the explanation

lib/finch.ex
keathley: Hmm..I think either are fine. I kinda like `:initial` since that describes how we're using that specific parameter

sneako: cool, I'll keep it as it!

test/finch_test.exs
sneako: Doh, you meant why did I use size & count instead of initial & max, that was a complete accident! Good news is that NimbleOptions made that clear immediately 🙃

PR #20, opened 2020-04-10
Telemetry

### LINKED ISSUE #4: Emit telemetry events
Finch should support telemetry events around requests. It probably makes sense to follow the new `span` conventions.

Here are the metrics that we tend to use when diagnosing production issues:

* queue duration
* new connection duration
* request duration
* response duration
* rate of new connections vs rate of reused connection

I think all of the durations we can get "for free" using spans.

keathley:
This PR adds telemetry events around our key events: queueing for a connection, connecting to the host, request, and response. I tried to keep things as clean as I could by moving most of the telemetry handling into its own module. I've also followed the current "span" conventions as described in #4. Before we move forward with this I think it would be good to run a benchmark in order to ensure we haven't accidentally added a bunch of overhead. I _think_ we should be OK. But it seems like something we should test.

If everything looks good I'll work on documenting the different events and their metadata in the `Telemetry` module. After that we should be g2g.

Closes #4


lib/finch/conn.ex
```diff
@@ -60,11 +70,21 @@ defmodule Finch.Conn do
...
     case HTTP.request(conn.mint, req.method, full_path, req.headers, req.body) do
       {:ok, mint, ref} ->
-        receive_response([], %{conn | mint: mint}, ref, %Response{}, receive_timeout)
+        Telemetry.stop(:request, start_time, metadata)
+        receive_response(%{conn | mint: mint}, ref, receive_timeout, metadata)
 
       {:error, mint, error} ->
+        metadata =
+          metadata
+          |> Map.put(:error, error)
```
sneako, line 86:
```suggestion
        metadata = Map.put(metadata, :error, error)
```

lib/finch/conn.ex
```diff
@@ -60,11 +70,21 @@ defmodule Finch.Conn do
   def request(conn, req, receive_timeout) do
     full_path = request_path(req)
 
+    metadata = %{
+      path: full_path,
+    }
```
sneako, line 75:
```suggestion
    metadata = %{path: full_path}
```

sneako:
So far this looks great!

lib/finch/telemetry.ex
```diff
@@ -0,0 +1,139 @@
...
+    measurements = %{duration: end_time - start_time}
+    meta =
+      meta
+      |> Map.put(:kind, kind)
+      |> Map.put(:error, reason)
+      |> Map.put(:stacktrace, stack)
+
+    :telemetry.execute([:finch, event, :exception], measurements, meta)
+  end
+
```
sneako, line 135:
```suggestion

  @doc false
```

sneako:
One thing that isn't clear to me at the moment is how to export spans from telemetry. Currently I am only exporting metrics, via statsd, and based on https://github.com/beam-telemetry/telemetry_metrics/issues/53 it seems that is still a work in progress. Maybe we should add {scheme, host, port} to the relevant events as well to make them more usable as metrics as well. WDYT?

keathley:
The "span" atm. is just convention for the `start`, `stop`, and `exception` events. If you just need statsd metrics you can use the `stop` and `exception` events to send histograms and counts. That's how we handle it in our services. In order to send out traces we attach to all 3 events in order to start, update and finish the different traces.

I think we can add scheme, host, and port. Should those be included on each metric?

sneako:
Yeah going over the list of events again, I think they could be useful on all of them.

keathley:
Sounds good :+1:

keathley:
Added the extra metadata and updated the docs.

PR #21, opened 2020-04-11
Change pools to be configured using URLs instead of shp tuples

### LINKED ISSUE #7: Allow pool configuration urls to be provided as a binary?
Right now we force the user to mentally parse the url in to the `{scheme, host, port}` (not that it's very difficult), but maybe some users would prefer to configure their pools like:
```
pools: %{
  "https://hex.pm" => %{size: 10, count: 2}
}
```
instead of the current:
```
pools: %{
  {:https, "hex.pm", 443} => %{size: 10, count: 2}
}
```

QuinnWilton:
This resolves #7 

This is mostly a straightforward change. The most controversial decision is making `parse_and_normalize_url/1` return `{parsed_uri, shp}`. The reason I did this is to ensure that URLs are always parsed and normalized consistently, when being used to construct an `shp` tuple, but without requiring two parses when `Finch.request/5` is called.

I also raise an `ArgumentError` now when an invalid scheme is passed to `normalize_scheme/1`, so that we can return better errors on pool startup.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -26,13 +26,14 @@ defmodule Finch do
   Finch.request(MyFinch, :get, "https://hex.pm")
   ```
 
-  When using HTTP/1, Finch will parse the passed in url into a `{scheme, host, port}`
+  When using HTTP/1, Finch will parse the passed in URL into a `{scheme, host, port}`
   tuple, and maintain one or more connection pools for each `{scheme, host, port}` you
   interact with.
 
-  You can also configure a pool size and count to be used for each specific
-  `{scheme, host, port}`s that are known before starting Finch. See `Finch.start_link/1` for
-  configuration options.
+  You can also configure a pool size and count to be used for specific URLs that are
+  known before starting Finch. The passed URLs will be parsed into `{scheme, host, port}`,
+  and the corresponding pools will be started. See `Finch.start_link/1` for configuration
+  options.
 
   ```elixir
   children = [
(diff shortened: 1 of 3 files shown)


lib/finch.ex
```diff
@@ -152,23 +154,31 @@ defmodule Finch do
...
     }
 
-    pool = PoolManager.get_pool(name, req.scheme, req.host, req.port)
+    pool = PoolManager.get_pool(name, shp)
     Pool.request(pool, req, opts)
   end
 
+  defp parse_and_normalize_url(url) when is_binary(url) do
+    uri = URI.parse(url)
+    uri = %URI{uri | scheme: normalize_scheme(uri.scheme), path: uri.path || "/"}
```
sneako, line 176: The type of `URI.scheme` is `nil | binary()` so setting it to an atom here breaks that contract. We should probably just build a map with the extracted info, and return that. 

I'm not really a fan of the tuple we are returning either. What do you think about returning ok or error tuples from this function? We can let the caller build a {s, h, p} tuple from the map returned on :ok, and on :error, we could let the error bubble up in `pool_options!/1`, where it will be raised and annotated with a bit more information. In `request/3,4,5` I think it would be better to return the error tuple and avoid raising. What do you think?

QuinnWilton: That sounds good to me! Nice catch with the scheme, too :)

QuinnWilton:
I pushed up changes to address your comments, and I think they make for a lot cleaner code. Thanks for the review!

Issue #22, opened 2020-04-11
Exponential backoff test is flaky


QuinnWilton:
On master (rarely):

```
  1) test request/5 will exponentially backoff until connection succeeds (FinchTest)
     test/finch_test.exs:205
     Assertion with >= failed
     code:  assert interval >= initial_backoff
     left:  35
     right: 37
     stacktrace:
       test/finch_test.exs:265: anonymous fn/3 in FinchTest."test request/5 will exponentially backoff until connection succeeds"/1
       (elixir 1.10.1) lib/enum.ex:783: Enum."-each/2-lists^foreach/1-0-"/2
       (elixir 1.10.1) lib/enum.ex:783: Enum.each/2
       test/finch_test.exs:262: (test)
```

keathley:
I think we'll probably have to add some padding to that test, or find a way to make it deterministic (controlling the clock for instance)

wojtekmach:
Alternatively I'd consider just testing the number of reconnections (which should give us confidence we're using backoff correctly) and not _when_ the reconnections happen - we assume that backoff does the right thing. On the flip side that would give us _less_ confidence in the system overall so it's definitely a tradeoff.

QuinnWilton:
I haven't looked into the issue in much detail, but is it possible that the issue is just that `backoff:rand_increment/2` explicitly allows for backoffs that are shorter than the previous backoff?

Here's the implementation: https://github.com/ferd/backoff/blob/master/src/backoff.erl#L42

I think we might be able to just check that the new backoff falls within that range.

wojtekmach:
it doesn't look like it can be shorter than the previous value but can be the same:

```elixir
iex> Stream.iterate(1, &:backoff.rand_increment(&1)) |> Enum.take(15)
[1, 1, 1, 1, 2, 3, 5, 5, 6, 7, 12, 33, 90, 132, 305]
```

QuinnWilton:
It can only happen with `:backoff.rand_increment/2`, when the max delay is "close" to the current delay.

Using possible numbers from the test:
```
iex(10)> Stream.iterate(50, &:backoff.rand_increment(&1, 100)) |> Enum.take(100) |> IO.inspect(charlists: :as_lists)
[50, 46, 96, 76, 42, 94, 49, 37, 54, 58, 97, 72, 83, 69, 55, 37, 50, 93, 87, 76,
[...]

Issue #23, opened 2020-04-12
Finch times out when attempting to connect to a bad address


mhanberg:
Currently when Finch attempts to connect to a server that doesn't exist/is down, it will time out instead of raising/throwing/returning an error stating that it can't connect.

I came across this when attempting to swap out HTTPoison with Finch in the [Wallaby](https://github.com/elixir-wallaby/wallaby) project.
![image](https://user-images.githubusercontent.com/5523984/79076630-a9f68e80-7cc9-11ea-9437-8b718f778d5a.png)

I think other HTTP clients will usually use an error along the lines of `:econnrefused` or `:nxdomain` for the corresponding connection errors.

mhanberg:
Might want to surface some some Mint.TransportErrors [here](https://github.com/keathley/finch/blob/master/lib/finch/conn.ex#L47).

keathley:
I think that might make sense. If we get an nxdomain error back there's probably not a lot we can do about that and should probably return that as an error to the client.

mhanberg:
At least concerning the transport errors, I think `:timeout`, might be the only one worth retrying.

keathley:
The current issue is that if we can't get a valid connection process from the pool, the pool will spin through all of the connections to try to find one that's valid. In the case where the service we're trying to communicate with is down, the pool will try to find a valid connection until the timeout is reached.

I think we'll have to update how we checkout a connection from the pool. We can either return a connection or an error. If we return an error the error can be surfaced to the caller. If we return a valid connection then the caller process will attempt to execute the request.

PR #24, opened 2020-04-19
Open connections lazily, in the client, rather than within the pool

sneako:
The current strategy works quite well if the server uses keep alive, however if the server does not, we see a significant performance degradation (even more than expected from the lack of keep alive).

If the server does not support keep alive or only allows a few requests per connection, all of our workers repeatedly have their connections closed, and reopened, which slows down our pool, and can also overwhelm the server. The problem gets worse with larger numbers of workers.

Instead, we can open the connections lazily, after giving it to the client. This gives the client better control over retries, and relieves pressure on the pool.

In my benchmarking, connecting lazily barely has an impact on throughput, if any at all, it only occurs when you first start using the pool, but once the connections have been established performance is essentially the same. 

As part of this change, I also removed the reconnections with exponential backoff which was needed for the async connect attempts, but now that we only connect on the caller, where we want to block as little as possible, I think it makes sense to remove, at least for now.

These changes would also correct the behaviour noted in https://github.com/keathley/finch/issues/23

### DIFF
--- a/lib/finch/conn.ex
+++ b/lib/finch/conn.ex
@@ -37,11 +37,11 @@ defmodule Finch.Conn do
     start_time = Telemetry.start(:connect, meta)
     conn_opts = Keyword.merge([mode: :passive], conn.opts)
 
-    with {:ok, mint} <- HTTP.connect(conn.scheme, conn.host, conn.port, conn_opts),
-         {:ok, mint} <- HTTP.controlling_process(mint, conn.parent) do
-      Telemetry.stop(:connect, start_time, meta)
-      {:ok, %{conn | mint: mint}}
-    else
+    case HTTP.connect(conn.scheme, conn.host, conn.port, conn_opts) do
+      {:ok, mint} ->
+        Telemetry.stop(:connect, start_time, meta)
+        {:ok, %{conn | mint: mint}}
+
       {:error, error} ->
         meta = Map.put(meta, :error, error)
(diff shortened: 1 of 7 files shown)


josevalim:
Looks great! Don't forget to run `mix deps.unlock --unused` too, if we have no other use for the backoff lib.

QuinnWilton:
This is awesome! It also resolves https://github.com/keathley/finch/issues/22

mhanberg:
Confirmed that it solves the issue was was seeing in #23 👍

lib/finch/conn.ex
```diff
@@ -25,32 +24,28 @@ defmodule Finch.Conn do
...
       host: conn.host,
       port: conn.port
     }
 
     start_time = Telemetry.start(:connect, meta)
+    conn_opts = Keyword.merge([mode: :passive], conn.opts)
 
-    with {:ok, mint} <- HTTP.connect(conn.scheme, conn.host, conn.port, conn.opts),
+    with {:ok, mint} <- HTTP.connect(conn.scheme, conn.host, conn.port, conn_opts),
          {:ok, mint} <- HTTP.controlling_process(mint, conn.parent) do
```
josevalim, line 41: WWe no longer need this line, right? Since we are now connecting from the process that is going to use it?

sneako: Ah good, point!

PR #31, opened 2020-04-22
upgrade to NimblePool master with precheckin/2

sneako:
Finishes the work started in https://github.com/keathley/finch/pull/30


josevalim:
:shipit:

PR #33, opened 2020-04-24
Allow Mint.connect/4 options to be configured

sneako:


lib/finch.ex
```diff
@@ -80,17 +80,25 @@ defmodule Finch do
...
     count: [
       type: :pos_integer,
       doc: "Number of pools to start.",
-      default: 1
+      default: @default_pool_count
+    ],
+    conn_opts: [
+      type: :keyword_list,
+      doc:
+        "These options are passed to [Mint.HTTP.connect/4](https://hexdocs.pm/mint/Mint.HTTP.html#connect/4-options) whenever a new connection is established. `:mode` is not configurable as Finch must control this setting. Typically these options are used to configure proxying, https settings, or connect timeouts.",
```
wojtekmach, line 100: nitpick: this is going to be slightly nicer for docs in IEx. Or linking is broken in nimble_options somehow?

```suggestion
        "These options are passed to `Mint.HTTP.connect/4` whenever a new connection is established. `:mode` is not configurable as Finch must control this setting. Typically these options are used to configure proxying, https settings, or connect timeouts.",
```

sneako: NIce! I didn't know that would work across libraries :)

wojtekmach: yup, but only if they are dependencies!

PR #34, opened 2020-04-24
Emit :handle_checkout events with idle_time measurements

### LINKED ISSUE #27: First Use Time & Idle Time Tracking
This will help in configuring pool size & count.

We can record System.monotonic_time on init and handle_checkin and subtract it from System.monotonic_time on handle_checkout.

Since init doesn't connect, maybe "first_use_time" makes sense as the name for that measurement.

sneako:
Closes #27 

The issue mentions possibly tracking "first_use_time" separately, but I think this measurement alone should suit our needs for now.

### DIFF
--- a/lib/finch/pool.ex
+++ b/lib/finch/pool.ex
@@ -34,8 +34,8 @@ defmodule Finch.Pool do
       NimblePool.checkout!(
         pool,
         :checkout,
-        fn from, conn ->
-          Telemetry.stop(:queue, start_time, metadata)
+        fn from, {conn, idle_time} ->
+          Telemetry.stop(:queue, start_time, metadata, %{idle_time: idle_time})
 
           with {:ok, conn} <- Conn.connect(conn),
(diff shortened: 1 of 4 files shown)


lib/finch/pool.ex
```diff
@@ -123,4 +120,19 @@ defmodule Finch.Pool do
       :closed
     end
   end
+
+  defp emit_checkout_event(conn) do
+    idle_time = System.monotonic_time() - conn.last_checkin
+    meta = telemetry_meta(conn, self())
+    Telemetry.event(:handle_checkout, %{idle_time: idle_time}, meta)
```
josevalim, line 127: Looks great! One alternative, so we emit fewer telemetry events, is to return the `idle_time` to the client, and include it as a measurement alongside `Telemetry.stop(:queue)`.

sneako: I like that idea! I was not totally sure that I should add a new event just for this.

PR #39, opened 2020-04-25
add docs for Finch.request/6

sneako:


lib/finch.ex
```diff
@@ -93,6 +123,18 @@ defmodule Finch do
     Supervisor.init(children, strategy: :one_for_all)
   end
 
+  @doc """
+  Sends an HTTP request and returns the response.
+
+  ## Options:
+    * `:pool_timeout` - (`pos_integer()` | `:infinity`) This timeout is applied when we check out
+    a connection from the pool.
```
wojtekmach, line 131: the built-in [`timeout/0`](https://hexdocs.pm/elixir/typespecs.html#built-in-types) is `non_neg_integer() | :infinity`, should it be like that too? Or, should we just skip it then?

I think we should also show default value.

```suggestion

    * `:pool_timeout` - This timeout is applied when we check out a connection from the pool.
      Default value is `5_000`.
```

lib/finch.ex
```diff
@@ -93,6 +123,18 @@ defmodule Finch do
...
+
+    * `:receive_timeout` - (`pos_integer()` | `:infinity`) The maximum time to wait for a response
+    before returning an error.
```
wojtekmach, line 134:
```suggestion
    * `:receive_timeout` - The maximum time to wait for a response before returning an error.
      Default value is `15_000`.
```

sneako:
Thanks @wojtekmach this is really great feedback! Clearly I don't have as much experience writing thorough documentation as you do 😁

README.md
```diff
@@ -2,7 +2,7 @@
 
 <!-- MDOC !-->
 
-An HTTP client with a focus on performance.
+An HTTP client with a focus on performance, built on top of `Mint` & `NimblePool`.
```
sneako, line 5: @wojtekmach when I generate the docs locally, NimblePool links correctly, but Mint does not. Any idea why?

lib/finch/response.ex
```diff
@@ -19,22 +19,9 @@ defmodule Finch.Response do
...
-  Headers are received as lists of two-element tuples containing two strings,
-  the header name and header value.
-  """
-  @type headers() :: [{header_name :: String.t(), header_value :: String.t()}]
-
   @type t :: %Response{
           status: status(),
-          body: body(),
-          headers: headers()
+          body: Finch.body(),
```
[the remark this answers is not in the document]
sneako: That makes a lot of sense. I have added separate types for each type of body 👍

README.md
wojtekmach: oh, that caught me by surprise too. The reason is there's no module `Mint`. (there's `Mint.HTTP`, etc)

sneako: oh of course, that does make sense. Hm I wonder if I should link to `Mint.HTTP` or use a normal markdown link to `https://hexdocs.pm/mint` instead 🤔

wojtekmach: yeah, both are fine, don't know which is more appropriate!

README.md
```diff
@@ -2,7 +2,7 @@
 
 <!-- MDOC !-->
 
-An HTTP client with a focus on performance.
+An HTTP client with a focus on performance, built on top of `Mint.HTTP` & `NimblePool`.
```
whatyouhide, line 5: My two cents to make things more discoverable :)

```suggestion
An HTTP client with a focus on performance, built on top of
[Mint](https://github.com/elixir-mint/mint) and [nimble_pool](https://github.com/dashbitco/nimble_pool).
```

lib/finch.ex
```diff
@@ -50,6 +50,28 @@ defmodule Finch do
...
 
+  @typedoc """
+  The `:name` provided to Finch in `start_link/1`.
+  """
+  @type name() :: atom()
+
+  @typedoc """
+  An HTTP request method represented as an atom or a String in all caps.
+
+  The following methods are supported: #{Enum.join(@methods, ", ")}.
```
whatyouhide, line 61: This makes me thing that any string methods are not supported, which is not true?

lib/finch.ex
```diff
@@ -50,6 +50,28 @@ defmodule Finch do
...
+
+  @typedoc """
+  A Uniform Resource Locator, the address of a resource on the Web.
+  """
+  @type url() :: String.t()
+
+  @typedoc """
+  A body associated with a request.
+  """
+  @type body() :: binary() | nil
```
whatyouhide, line 73: Just curious, but doesn't Finch support iodata?

sneako: Ah right, it certainly does, thanks!

lib/finch.ex
sneako: Yes, in the current state, arbitrary string methods are not supported.

```
iex(2)> Finch.request(MyFinch, "WAT", "https://appmonet.com")
** (FunctionClauseError) no function clause matching in Finch.build_method/1

    The following arguments were given to Finch.build_method/1:

        # 1
        "WAT"

    Attempted function clauses (showing 2 out of 2):

        defp build_method(method) when method === "GET" or (method === "POST" or (method === "PUT" or (method === "PATCH" or (method === "DELETE" or (method === "HEAD" or method === "OPTIONS")))))
        defp build_method(method) when is_atom(method)

```

But I can't think of any good reason not to allow that flexibility, so I will open an issue to address that.

README.md
```diff
@@ -2,7 +2,8 @@
 
 <!-- MDOC !-->
 
-An HTTP client with a focus on performance.
+An HTTP client with a focus on performance, built on top of
+[Mint](https://github.com/elixir-mint/mint) and [nimble_pool](https://github.com/dashbitco/nimble_pool).
```
wojtekmach, line 6:
```suggestion
[Mint](https://github.com/elixir-mint/mint) and [NimblePool](https://github.com/dashbitco/nimble_pool).
```

PR #41, opened 2020-04-25
Allow arbitrary string methods to be used

### LINKED ISSUE #40: Support arbitrary string methods
Right now, we currently have guard clauses that are perhaps a bit too restrictive and therefore do not allow the user to use arbitrary HTTP methods.

Are there any good reasons to keep this limitation, or should it be removed?

sneako:
Closes #40 

Unfortunately, I don't think I can write a test for non standard HTTP methods using bypass, as the [Plug.Router](https://github.com/elixir-plug/plug/blob/8efb591f53c87c51a9aded8212daa78e7c3ff83e/lib/plug/router.ex#L47) only supports the standard ones.

Either way, we will need a new test server once we introduce HTTP/2 support, so I will begin looking in to a custom HTTP/1 test server now as well.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -131,15 +131,14 @@ defmodule Finch do
   end
 
   defp build_method(method) when is_binary(method), do: method
-
-  defp build_method(method) when is_atom(method) and method in @atom_methods do
-    @atom_to_method[method]
-  end
+  defp build_method(method) when method in @atom_methods, do: @atom_to_method[method]
 
   defp build_method(method) do
     raise ArgumentError, """
     got unsupported atom method #{inspect(method)}.
-    only the following methods can be provided as atoms: #{Enum.map_join(@atom_methods, ", ", &inspect/1)}",
+    only the following methods can be provided as atoms: #{
+      Enum.map_join(@atom_methods, ", ", &inspect/1)
+    }",
     otherwise you must pass a binary.
     """
   end
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -130,10 +130,18 @@ defmodule Finch do
     end
   end
 
-  defp build_method(method) when method in @methods, do: method
+  defp build_method(method) when is_binary(method), do: method
 
-  defp build_method(method) when is_atom(method) do
+  defp build_method(method) when is_atom(method) and method in @atom_methods do
     @atom_to_method[method]
+  end
+
+  defp build_method(method) do
+    raise ArgumentError, """
+    got unsupported atom method #{inspect(method)}.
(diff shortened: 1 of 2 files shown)


lib/finch.ex
```diff
@@ -130,12 +130,20 @@ defmodule Finch do
     end
   end
 
-  defp build_method(method) when method in @methods, do: method
+  defp build_method(method) when is_binary(method), do: method
 
-  defp build_method(method) when is_atom(method) do
+  defp build_method(method) when is_atom(method) and method in @atom_methods do
     @atom_to_method[method]
```
wojtekmach, line 136: I think this can now become simply:

```suggestion
  defp build_method(method) when method in @atom_methods do
    method |> Atom.to_string() |> String.upcase()
```

and you can remove `@methods` and `@atom_to_method`.

lib/finch.ex
```diff
@@ -130,12 +130,20 @@ defmodule Finch do
...
   end
 
+  defp build_method(method) do
+    raise ArgumentError, """
+    got unsupported atom method #{inspect(method)}.
+    only the following methods can be provided as atoms: #{Enum.join(@atom_methods, ", ")}",
```
wojtekmach, line 142:
```suggestion
    only the following methods can be provided as atoms: #{Enum.map_join(@atom_methods, ", ", &inspect/1)}",
```

test/finch_test.exs
```diff
@@ -185,6 +185,13 @@ defmodule FinchTest do
                end)
     end
 
+    test "raises if unsupported atom request method provided", %{bypass: bypass} do
+      error =
+        assert_raise(ArgumentError, fn -> Finch.request(MyFinch, :gimme, endpoint(bypass)) end)
+
+      assert error.message =~ "got unsupported atom method :gimme"
```
wojtekmach, line 192:
```suggestion
      assert_raise ArgumentError, ~r/got unsupported atom method :gimme/, fn ->
        Finch.request(MyFinch, :gimme, endpoint(bypass))
      end
```

lib/finch.ex
sneako: Ah right, I can definitely remove the is_atom/1 guard, but wouldnt the map lookup be slightly faster than the conversion?

sneako: I did a quick benchmark and got these results:
```
Name                 ips        average  deviation         median         99th %
map access        8.81 M      113.53 ns   ±409.74%           0 ns        1000 ns
convert           3.44 M      290.47 ns  ±6416.53%           0 ns        1000 ns

Comparison:
map access        8.81 M
convert           3.44 M - 2.56x slower +176.95 ns
```

PR #44, opened 2020-04-28
fix typespec

bismark:
Causes dialyzer errors and conflicts with docs.


sneako:
Ah right, the spec is missing `:put`. Nice catch, thank you!

PR #45, opened 2020-04-29
Round robin pool selection

sneako:
I used the `:atomics` module, so that means we would no longer support OTP versions < 21.2. We could fallback to an :ets counter if we feel that supporting older OTP versions is necessary. Or we could possibly just not allow this strategy in older versions


lib/finch/pool/strategy.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Finch.Pool.Strategy do
```
sneako, line 1: These new Pool modules should probably be private/@moduledoc false, but i felt the documentation could be useful when implementing additional strategies

lib/finch/pool.ex
```diff
@@ -55,42 +61,42 @@ defmodule Finch.Pool do
   end
 
   @impl NimblePool
-  def init_pool({registry, shp, opts}) do
-    {:ok, _} = Registry.register(registry, shp, [])
-    {:ok, {shp, opts}}
+  def init_pool({registry, state}) do
+    {:ok, _} = Registry.register(registry, state.shp, state.registry_value)
```
sneako, line 65: I realize that we will have this registry value duplicated for every single pool, but I think this is the best way to quickly look it up at the same time as the pool pid.

lib/finch/pool/round_robin.ex
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  @impl true
+  def handle_enqueue(%{counter: counter}) do
+    :counters.add(counter, 1, 1)
+  end
+
+  @impl true
+  def choose_pool([{_, %{counter: counter, count: count}} | _] = pids) do
+    index = :counters.get(counter, 1)
```
josevalim, line 20: Unfortunately I think this does not really guarantee round robin, because you can have handle_enqueue being called multiple times before you get to choose the pool. Or you choose the pool multiple times before handle_enqueue is called and bumped. You think you would need to use ets:update_counter if you want an actual round robin.

sneako: Ahh right of course :( Ok, I'll switch to the ets counter. Thanks for pointing that out!

lib/finch.ex
```diff
@@ -107,12 +121,16 @@ defmodule Finch do
...
+    ets_name = ets_name(config.registry_name)
+    config = Map.put(config, :ets_name, ets_name)
+
     children = [
       {DynamicSupervisor, name: config.supervisor_name, strategy: :one_for_one},
       {Registry, [keys: :duplicate, name: config.registry_name, meta: [config: config]]},
       {PoolManager, config}
     ]
 
+    :ets.new(ets_name, [:public, :named_table])
```
sneako, line 133: Currently investigating just using the table reference (and not naming the table) and if :write_concurrency & :read_concurrency have any significant impact.

lib/finch/pool/round_robin.ex
michalmuskala: You can probably use the `atomics` module that has a `add_get` function

sneako: That's right! I try that out

lib/finch/pool/selection_strategy/round_robin.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Finch.Pool.SelectionStrategy.RoundRobin do
+  @moduledoc """
+  Cycle through the pools one by one, in a consistent order.
+  """
+  @behaviour Finch.Pool.SelectionStrategy
+
+  @impl true
+  def registry_value(%{count: count}) do
+    atomics = :atomics.new(1, [])
```
IRog, line 9: I think you might want this to be unsigned. I think you're relying on overflow behavior and this is going into undefined behavior as it's eventually a signed-int overflow. I could be wrong.

sneako: Great catch! I am relying on the overflow behaviour, but you're right that if the value becomes negative, things will break. Thank you!

michalmuskala: AFAIK `atomics` use intrinsics that guarantee proper wraparound for signed integers as well

sneako:
I did not see any noticeable improvement from these changes in my benchmarks so I left this PR sitting around for that reason. I am definitely still interested in looking in to alternative pool selection strategies, but I will close this PR because it is so outdated.

PR #47, opened 2020-05-01
Accepts a URI struct in request/3/4/5/6

### LINKED ISSUE #46: accept %URI{} as well as binary for url()
https://hexdocs.pm/finch/Finch.html#t:url/0 defines a url as a binary(). I am accustomed to passing around %URI{} structs instead.

Is there a reason you wouldn't also accept (or perhaps only expect) a %URI{} struct?

BTW - love the library, just thought I'd ask.

supersimple:
Closes #46

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -66,7 +66,7 @@ defmodule Finch do
   @typedoc """
   A Uniform Resource Locator, the address of a resource on the Web.
   """
-  @type url() :: String.t() | %URI{}
+  @type url() :: String.t() | URI.t()
 
   @typedoc """
   A body associated with a request.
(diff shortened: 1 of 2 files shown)


lib/finch.ex
```diff
@@ -66,7 +66,7 @@ defmodule Finch do
   @typedoc """
   A Uniform Resource Locator, the address of a resource on the Web.
   """
-  @type url() :: String.t()
+  @type url() :: String.t() | %URI{}
```
wojtekmach, line 69:
```suggestion
  @type url() :: String.t() | URI.t()
```

keathley:
Thanks @supersimple :tada:

PR #49, opened 2020-05-05
Build abstraction around pools

keathley:
This PR introduces an abstraction around the pool. This will allow us to choose between either http1 or http2 pools in the future.

The main idea here is that we store the pool module as the "value" in the registry. When we request a pool connection we use the provided module. For http1 pools this is a nimble pool which will hand the caller a specific connection. For http2 pools the "pool" can be a process that holds a single http2 connection.

There are still some vestigial bits for http2 left in this PR. I'm not sure if we want to remove those now or if its not a big deal. We'll need most of this stuff when we actually implement the http2 connection process. But if its cleaner to remove it I can do that.


lib/finch.ex
```diff
@@ -32,6 +32,11 @@ defmodule Finch do
   @default_pool_count 1
 
   @pool_config_schema [
+    scheme: [
```
sneako, line 35: Looks great! The only thing I am not sure about is calling this `:scheme`, since we already use scheme throughout the codebase to refer to `:http` or `:https`, so it could cause some confusion. Maybe we could call it `:protocol`?

keathley: Yup. That makes way more sense.

PR #50, opened 2020-05-06
Return empty string for missing response bodies

keathley:
So, after using starting to roll this out into production I've received a bunch of feedback that we should default the response body to an empty string. After thinking about it more I think that is probably the correct choice for a few reasons:

* It's not "wrong" semantically speaking. This is the most important point IMO. From the point of view of the parser, we'd expect it to return something like an empty string.
* It removes additional error checking logic that users need to do.
* There's a precedent for this in elixir already. The other clients return empty strings.
* Clients in other languages also use an empty string. I had only glanced at the ruby client. At this point, I'm pretty sure ruby is the outlier here.

That's my reasoning for making this change now.

TL;DR - @QuinnWilton was right :heart:

### DIFF
--- a/lib/finch/response.ex
+++ b/lib/finch/response.ex
@@ -7,18 +7,13 @@ defmodule Finch.Response do
 
   defstruct [
     :status,
-    :body,
+    body: "",
     headers: []
   ]
 
-  @typedoc """
-  A body associated with a response.
-  """
-  @type body() :: binary() | nil
-
   @type t :: %Response{
           status: Mint.Types.status(),
-          body: body(),
+          body: binary(),
           headers: Mint.Types.headers()
         }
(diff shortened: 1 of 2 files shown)


lib/finch/response.ex
```diff
@@ -7,14 +7,14 @@ defmodule Finch.Response do
...
-    :body,
+    body: "",
     headers: []
   ]
 
   @typedoc """
   A body associated with a response.
   """
-  @type body() :: binary() | nil
+  @type body() :: binary()
```
wojtekmach, line 17: consider just inlining this type: `%Response{..., body: binary(), ...}`. perhaps it's useful to export it, idk, so please feel free to just ignore me. :)

PR #53, opened 2020-05-09
HTTP2 pool support

keathley:
I hope you don't mind @whatyouhide but I kinda wanted to give this a try. Mostly because I wanted to make sure I understood exactly what was going on here. This is based on the work that y'all had done in mint_pool. The only real change is the removal of the explicit streaming options.

I've added some very basic tests in here. But we should work on testing more of the failure conditions and request timeouts. I think we're also going to have to work out what to do if we already have too many inflight requests for a given pool. Its possible that we can buffer a small number of requests. But that buffer would have to be bounded. It also means that we would need to look at implementing something like CoDel in order to evict requests from the buffer. I think both of these are reasonable and make sense to add to Finch. For now we're just going to drop the requests, which is a safer option for users.

Lemme know what you think.


whatyouhide:
> I hope you don't mind @whatyouhide

I don't :)

> The only real change is the removal of the explicit streaming options.

Why?

> Its possible that we can buffer a small number of requests. But that buffer would have to be bounded. It also means that we would need to look at implementing something like CoDel in order to evict requests from the buffer. I think both of these are reasonable and make sense to add to Finch. For now we're just going to drop the requests, which is a safer option for users.

Make it work, make it beautiful 😛 I definitely think this is a "post MVP" feature.


You could take a bunch of integration tests from mint_pool for this. At least they cover _something_ :)

keathley:
@whatyouhide I dropped the explicit streaming function since there's no top-level `stream_request` type API for Finch. We'd need to make sure that that would work with http 1.1 requests as well before rolling it out. We haven't needed to do that yet but we can always investigate in the future. For now the http/2 pool only exposes a `request` function which collects the full response before returning to the user.

sneako:
Really excited to try this out! @keathley how about the telemetry support? Were you thinking that should be added in a later PR?

keathley:
Yup. I’d like to add telemetry and some more tests before we merge this in. We should probably do some benchmarking as well just to see where we stand

keathley:
@sneako Updated with telemetry and more tests.

lib/finch/http2/pool.ex
```diff
@@ -0,0 +1,409 @@
...
+      {:ok, conn} ->
+        Telemetry.stop(:connect, start, metadata)
+        data = %{data | conn: conn}
+        {:next_state, :connected, data}
+
+      {:error, error} ->
+        metadata = Map.put(metadata, :error, error)
+        Telemetry.stop(:connect, start, metadata)
+        Logger.error([
+          "Failed to connect to #{data.scheme}:#{data.host}:#{data.port}: ",
```
sneako, line 160:
```suggestion
          "Failed to connect to #{data.scheme}://#{data.host}:#{data.port}: ",
```

lib/finch.ex
```diff
@@ -147,6 +147,9 @@ defmodule Finch do
         query: uri.query
       }
 
+      # TODO - Use Nimble options to build teh request opts
```
sneako, line 150:
```suggestion
      # TODO - Use Nimble options to build the request opts
```

lib/finch/http2/pool.ex
```diff
@@ -0,0 +1,409 @@
...
+
+        if HTTP2.open?(conn, :read) do
+          {:keep_state, data, actions}
+        else
+          {:next_state, :disconnected, data, actions}
+        end
+
+      {:error, conn, error, responses} ->
+        Logger.error([
+          "Received error from server #{data.scheme}:#{data.host}:#{data.port}: ",
```
sneako, line 324:
```suggestion
          "Received error from server #{data.scheme}://#{data.host}:#{data.port}: ",
```

test/finch/http2/pool_test.exs
```diff
@@ -0,0 +1,281 @@
...
+    {:ok, pool} =
+      start_server_and_connect_with([server_settings: server_settings], fn port ->
+        start_pool(port)
+      end)
+
+    assert {:error, error} = Pool.request(pool, %{req | headers: [{"foo", "bar"}]}, [])
+    assert %{reason: {:max_header_list_size_exceeded, _, _}} = error
+  end
+
+  @tag :capture_log
```
sneako, line 71:
```suggestion
```
Since we have the `@moduletag`

test/finch/http2/pool_test.exs
```diff
@@ -0,0 +1,281 @@
...
+    # If the server now closes the socket, we actually shut down.
+    :ok = :ssl.close(server_socket())
+
+    Process.sleep(50)
+
+    # If we try to make a request now that the server shut down, we get an error.
+    assert {:error, %{reason: :disconnected}} = Pool.request(pool, req, [])
+  end
+
+  @tag :capture_log
```
sneako, line 108:
```suggestion
```

PR #59, opened 2020-06-09
Add streaming support

josevalim:
Both HTTP1 and HTTP2 already streamed internally,
this PR simply moves the stream out.

This PR also changed the main Finch API to have a
explicit step that builds the request. Instead of:

    Finch.request(MyApp, :get, "/")

You must now do:

    Finch.build(:get, "/") |> Finch.request(MyApp)

This is necessary otherwise the `Finch.stream/5`
API would be very complex. One advantage of this
approach is also that we no longer need to pass
empty headers when configuring get requests, for
example, instead of:

    Finch.request(MyApp, :get, "/", [], nil, pool_timeout: 10_000)

One can now write:

    Finch.build(:get, "/") |> Finch.request(MyApp, pool_timeout: 10_000)

The previous API has been properly deprecated.

Finally, this PR streamlines the `receive_timeout`
value. Previously the documentation and each pool
used a different value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Once you have started Finch, you can use the client you have started,
 by passing the name of your client as the first argument of `Finch.request/3,4,5,6`:
 
 ```elixir
-Finch.request(MyFinch, :get, "https://hex.pm")
+Finch.build(:get, "https://hex.pm") |> Finch.request(MyFinch)
 ```
 
 When using HTTP/1, Finch will parse the passed in URL into a `{scheme, host, port}`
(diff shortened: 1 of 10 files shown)


keathley:
I've looked through this and everything looks fine to me codewise. I'm going back and forth on the API change. I get the reason to do it and it's probably the best option. I just need to get comfortable with it ;)

sneako:
Very cool! Thank you @josevalim! I'll run the benchmarks at some point today just to confirm there is no significant performance regression.

keathley:
Ok, looking over it some more, I'm good with this new API and it should allow for more flexibility in the future. I'm good to merge this unless you wanna do benchmarks first @sneako. I can't imagine there's any substantial perf hit from this.

sneako:
I'm happy to merge as well, thanks again @josevalim !

PR #61, opened 2020-06-09
Update README.md

garthk:
The Finch name isn't the first argument in the new API.


sneako:
Thanks, good catch! However, the update is a bit misleading, because Finch does not have to be started in order for you to build your request. For example, I am playing around with building the request at compile time, and sticking it in a module attribute.

PR #66, opened 2020-06-26
Respect connection close header

josevalim:


josevalim:
Mint already closes it. :) Thanks for verifying @sneako !

PR #67, opened 2020-07-02
Fix finch_test - protocol_not_negotiated

xinz:
Hello,

When upgrade Mint from 1.0.0 to 1.1.0, run Finch's tests will get the following error:

```
1) test connection options are passed through to the conn (FinchTest)
     test/finch_test.exs:271
     match (=) failed
     code:  assert {:error, _} = Finch.build(:get, endpoint(bypass)) |> Finch.request(H2Finch)
     left:  {:error, _}
     right: {:ok, %Finch.Response{body: "OK", headers: [{"cache-control", "max-age=0, private, must-revalidate"}, {"content-length", "2"}, {"date", "Thu, 02 Jul 2020 09:06:35 GMT"}, {"server", "Cowboy"}], status: 200}}
     stacktrace:
       test/finch_test.exs:283: anonymous fn/1 in FinchTest."test connection options are passed through to the conn"/1
       (ex_unit 1.10.3) lib/ex_unit/capture_log.ex:74: ExUnit.CaptureLog.capture_log/2
       test/finch_test.exs:282: (test)
```

Since Mint's [commit](https://github.com/elixir-mint/mint/commit/4882d9d8090fa3fd7e45b27141ec60fcf1d488ad) released in 1.1.0 will not support protocol negotiation for TCP connections, Finch's original test case will be break after this upgrade.

I refer Mint's similar [test case](https://github.com/elixir-mint/mint/blob/master/test/mint/integration_test.exs#L32) to modify the above fail test case, its works fine with Mint 1.0.0 and 1.1.0, please review.

Thanks,
Xin

### DIFF
--- a/test/finch_test.exs
+++ b/test/finch_test.exs
@@ -278,12 +278,12 @@ defmodule FinchTest do
       stop_supervised(Finch)
     end
 
-    test "SSL - fail to select HTTP2", _bypass do
+    test "fail to negotiate h2 protocol", _bypass do
       start_supervised!({Finch, name: H2Finch, pools: %{default: [protocol: :http2]}})
 
       capture_log(fn ->
         assert {:error, _} = Finch.build(:get, "https://httpstat.us") |> Finch.request(H2Finch)
-      end) =~ "failed to negotiate protocol"
+      end) =~ "ALPN protocol not negotiated"
     end
 
     test "caller is unable to override mode", %{bypass: bypass} do
--- a/test/finch_test.exs
+++ b/test/finch_test.exs
@@ -276,11 +276,13 @@ defmodule FinchTest do
       assert {:ok, _} = Finch.build(:get, endpoint(bypass)) |> Finch.request(H1Finch)
 
       stop_supervised(Finch)
+    end
 
+    test "SSL - fail to select HTTP2", _bypass do
       start_supervised!({Finch, name: H2Finch, pools: %{default: [protocol: :http2]}})
 
       capture_log(fn ->
-        assert {:error, _} = Finch.build(:get, endpoint(bypass)) |> Finch.request(H2Finch)
+        assert {:error, _} = Finch.build(:get, "https://httpstat.us") |> Finch.request(H2Finch)
       end) =~ "failed to negotiate protocol"
     end
 
--- a/test/finch_test.exs
+++ b/test/finch_test.exs
@@ -1,6 +1,5 @@
 defmodule FinchTest do
   use ExUnit.Case, async: true
-  import ExUnit.CaptureLog
   doctest Finch
 
   alias Finch.Response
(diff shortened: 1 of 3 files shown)


test/finch_test.exs
```diff
@@ -276,11 +276,13 @@ defmodule FinchTest do
       assert {:ok, _} = Finch.build(:get, endpoint(bypass)) |> Finch.request(H1Finch)
 
       stop_supervised(Finch)
+    end
 
+    test "SSL - fail to select HTTP2", _bypass do
       start_supervised!({Finch, name: H2Finch, pools: %{default: [protocol: :http2]}})
 
       capture_log(fn ->
```
keathley, line 284: This doesn't actually test anything because there's no assert. The actual log returned here is `ALPN protocol not negotiated`.

xinz: Thanks for your reply @keathley, I have some update to make this test case more accurate (I hope so), please review and let me know if others need to tweak to keep us in the positive.

xinz:
Thanks, I understand your intention now, in my opinion if the external service is stable and trusted that would take more testing cases, and can reduce our local test HTTP server related working, but there are many trade-off need to evaluate as well, I am open about this.

test/finch_test.exs
xinz: Sorry that the missed `assert` is added now, please review again.

xinz:
FYI, I try to setup local HTTP/1.1 server with the self-signed certs, here are my local sample:

```
defmodule Finch.HTTP1Server do
  @moduledoc false

  @fixtures_dir Path.expand("../fixtures", __DIR__)

  def start(port) do
    children = [
      Plug.Adapters.Cowboy.child_spec(
        scheme: :https,
        plug: Finch.HTTP1Server.PlugRouter,
        options: [
          port: port,
          cipher_suite: :strong,
          certfile: Path.join([@fixtures_dir, "selfsigned.pem"]),
          keyfile: Path.join([@fixtures_dir, "selfsigned_key.pem"]),
          alpn_preferred_protocols: ["http/1.1"],
          next_protocols_advertised: ["http/1.1"],
          otp_app: :finch,
          protocol_options: [
            idle_timeout: 3_000,
            inactivity_timeout: 5_000,
            max_keepalive: 1_000,
            request_timeout: 10_000,
            shutdown_timeout: 10_000
          ]
        ]
      )
    ]

    Supervisor.start_link(children, strategy: :one_for_one)
  end
end
```

But I occur this error for connection failed, and I do not find the right way to figure out this issue, if anyone can help me on this, we may remove the external service if possible, I feel this local setup increased complexity for me, but I gain new knowledge from this.

Thank you very much!

```
Failed to connect to https://localhost:4002: TLS client: In state hello received SERVER ALERT: Fatal - No application protocol
```

xinz:
>  I also think that relying on Mint's logging for these tests is going to be pretty brittle. As a user of the API, I would probably expect to get some sort of error back.

Agree, we should use the matched error instead of the logging in test cases, I am learning Finch, and I have some questions about the `return error` need yours help, I will open another issue to listen to yours opinions.

QuinnWilton:
> I'm not a huge fan of relying on an external service in these tests but perhaps that makes the most sense?

One approach I sometimes take for cases like this is tagging the test as requiring an external service, disabling those tests as part of the automated CI, but running them by default otherwise. It isn't perfect, but it keeps them being run by people locally without holding up merges.

xinz:
FYI, after some investigations, looks like Cowboy does not support disable HTTP/2 so far, here is my finding from its [rfc7540_SUITE](https://github.com/ninenines/cowboy/blob/master/test/rfc7540_SUITE.erl#L856) test, please correct me if any missing.

xinz:
@keathley hello, we have initialized an http/1.1 integration test to replace the external service, could you please review these changes to keep the testing pass with `{:mint, "~> 1.1"}`, and then we can update `assert capture_log` in the test to use returned error in the assertion later if it is acceptable.

Please let me know if any I can help on it.

keathley:
Thanks @xinz, this has been merged into main.

PR #68, opened 2020-07-06
Include http method in req and resp metrics

cloud8421:
So that metrics can be grouped by method where possible.

The method would be an uppercased string, which I think is appropriate for the use case.


keathley:
Thanks @cloud8421! :tada:

GregMefford:
Yay! I've been meaning to carve out some time to add this for a while now. 🚀

PR #72, opened 2020-07-14
Try addressing remaining closed connection issues

josevalim:
Now only the peerdata one.

/cc @sneako


sneako:
This has not affected the rate of connection closed errors, and I am also seeing the port leak that we fixed with NimblePool.precheckin return, so I have opened a PR to put back that function call https://github.com/keathley/finch/pull/73

sneako:
Hmm, I am actually still seeing the port leak with the precheckin in place... We may need to revert #73 and #71

PR #73, opened 2020-07-14
put back NimblePool.checkin

sneako:
We added the precheckin functionality to NimblePool to address a port leak here: https://github.com/keathley/finch/pull/31 and it was removed here: https://github.com/keathley/finch/pull/71

Unfortunately, without the precheckin, the port leak has returned

/cc @josevalim


josevalim:
I think this doesn't fix all leaks because we are transferring when we connect, before we do the request, so there is a gap between connect and transfer. I will think about it. :)

sneako:
That makes sense :( I will revert this and #71 for now and make sure I run longer tests for port leaks going forward

PR #76, opened 2020-07-14
Another attempt at reducing transfer

josevalim:
This should fix the port leak too, but please verify.


sneako:
Benchmarks look good, only slightly slower than the previous pr, but still faster than master! I'll check for port leaks now

sneako:
Port count is holding steady!

sneako:
I let this run in production overnight and everything looks good 👍

PR #77, opened 2020-07-16
HTTP/1 close the connection if the transfer fails

sneako:
Right now we ignore the return value of `Conn.transfer/2` when we transfer fresh connections to the pool. This is ok, because `Conn.set_mode/2` will fail in `handle_checkin/3`, and the connection will be removed at this point, but with this change, we can skip the attempt to set_mode in cases when we know it will fail.

I don't think this will fix the connection closed error we are tracking down, but this is a bit more efficient than current master in my benchmarks.

/cc @josevalim


sneako:
I have been running this in production for a few hours now, looking good 👍

PR #78, opened 2020-08-25
Include query params in http2 requests

bismark:


sneako:
Ah wow, thanks for noticing this! I am still only using the http1 pool so I haven't come across this yet. Would you mind adding a test case to prove that the query string is sent with the request?

sneako:
Thanks again @bismark ! Closing in favor of https://github.com/keathley/finch/pull/79

PR #87, opened 2020-09-29
Use HTTP1 in HTTP1 pools

josevalim:
Otherwise Mint may try HTTP2 if available.

### DIFF
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -1,7 +1,7 @@
 defmodule Finch.Conn do
   @moduledoc false
 
-  alias Mint.HTTP
+  alias Mint.HTTP1
   alias Finch.Telemetry
 
   def new(scheme, host, port, opts, parent) do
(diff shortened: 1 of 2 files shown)


sneako (approved):

lib/finch/http1/conn.ex
```diff
@@ -50,29 +50,31 @@ defmodule Finch.Conn do
...
   def set_mode(conn, mode) when mode in [:active, :passive] do
-    case HTTP.set_mode(conn.mint, mode) do
+    case HTTP1.set_mode(conn.mint, mode) do
       {:ok, mint} -> {:ok, %{conn | mint: mint}}
       _ -> {:error, "Connection is dead"}
     end
   end
 
-  def stream(%{mint: nil}, _), do: {:error, "Connection is dead"}
+  def discard(%{mint: nil}, _), do: :unknown
```
josevalim, line 71: If there is no connection, then the message is certainly `:unknown` for this particular worker.

keathley:
This has been merged into main.

PR #92, opened 2020-10-08
Introduce max_idle_time for HTTP1 connections

drewolson:
@keathley as we discussed yesterday on IRC, this PR adds an optional `max_idle_time` pool configuration. This change allows us to deal with misbehaving endpoints that do not close connections cleanly. Before this change, we were seeing `{:error, %Mint.TransportError{reason: :closed}}` when Finch attempted to reuse connections that had not been cleanly closed.

After this commit, the user can specify a `max_idle_time` in milliseconds and, when checking out an existing connection, the pool will verify that the current idle time is less than this specified time. If it is greater, the pool will close the Mint connection and remove the worker from the pool.

This change should be backwards compatible as the default `max_idle_time` is `:infinity`.

We weren't sure how to add ex_unit tests for this change, nor if this option should have any effect on the HTTP2 pool. However, we have verified that deploying this change into our QA environment does indeed solve the problem we were previously seeing.

We also tested the change locally via iptables. To induce this misbehavior, you can run any kind of simple HTTP server locally (let's suppose it's on port 4000) and use this type of iptables rule to drop FIN packets from that server, which simulates a remote endpoint that does not cleanly close connections, leaving a stale connection in the Finch pool.

```
iptables -A INPUT --protocol tcp --source-port 4000 --tcp-flags FIN FIN -j DROP
```

Steps to reproduce:

1. Run a test HTTP endpoint locally, and use iptables to block FIN packets from that server (see above)
2. Configure Finch with a pool size of 1 to guarantee we hit the stale connection in the next steps
3. Make one request to the test HTTP server through Finch
4. Wait 60+ seconds for the remote endpoint to time out the idle connection (you can verify a stream of FINs are attempted to be sent by the server via tcpdump, etc.)
5. Make a new request through the Finch pool. You will see `{:error, %Mint.TransportError{reason: :closed}}`.

Thanks!
Drew and Mike

Signed-off-by: Mike Pilat <mike.pilat@greensill.com>

### DIFF
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -12,6 +12,7 @@ defmodule Finch.Conn do
       opts: opts.conn_opts,
       parent: parent,
       last_checkin: System.monotonic_time(),
+      max_idle_time: opts.max_idle_time,
       mint: nil
     }
   end
@@ -60,6 +61,16 @@ defmodule Finch.Conn do
 
   def open?(%{mint: nil}), do: false
   def open?(%{mint: mint}), do: HTTP1.open?(mint)
+
+  def idle_time(conn, unit \\ :native) do
+    idle_time = System.monotonic_time() - conn.last_checkin
+
+    System.convert_time_unit(idle_time, :native, unit)
+  end
+
+  def reusable?(conn) do
+    idle_time(conn, :millisecond) < conn.max_idle_time
+  end
 
   def set_mode(conn, mode) when mode in [:active, :passive] do
     case HTTP1.set_mode(conn.mint, mode) do
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -68,6 +68,8 @@ defmodule Finch.Conn do
     System.convert_time_unit(idle_time, :native, unit)
   end
 
+  def reusable?(%{max_idle_time: :infinity}), do: true
+
   def reusable?(conn) do
     idle_time(conn, :millisecond) < conn.max_idle_time
   end
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -28,6 +28,12 @@ defmodule Finch do
       doc: "Number of pools to start.",
       default: @default_pool_count
     ],
+    max_idle_time: [
+      type: :timeout,
+      doc:
+        "The maxiumum number of milliseconds an HTTP1 connection is allowed to be idle before being closed during a checkout attempt",
+      default: :infinity
+    ],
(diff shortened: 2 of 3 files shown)


lib/finch/http1/conn.ex
```diff
@@ -61,6 +62,16 @@ defmodule Finch.Conn do
   def open?(%{mint: nil}), do: false
   def open?(%{mint: mint}), do: HTTP1.open?(mint)
 
+  def idle_time(conn, unit \\ :native) do
+    idle_time = System.monotonic_time() - conn.last_checkin
+
+    System.convert_time_unit(idle_time, :native, unit)
+  end
+
+  def reusable?(conn) do
```
sneako, line 71: I like this idea a lot, thanks for the contribution!

Maybe we should add a clause that would skip the check if max_idle_time is infinity?

drewolson: Currently, `:infinity` "just works" with `<` (I believe this is the reason it is idiomatic to use it for `timestamp`s and allowed by nimble options).

That said, if folks think it would be more intentional to explicitly skip the check when the value of `:infinity` is present, I'd be ok doing so.

sneako: I'm sure that it works, but I was thinking of it more as a minor optimization, to skip the work done in `idle_time/2`

drewolson:
Ah, makes sense. I’m happy to make the update.

lib/finch/http1/conn.ex
drewolson: Whoops, my email reply was out of the thread. Happy to make this update.

drewolson: Fixed in https://github.com/keathley/finch/pull/92/commits/c4644d1a4bc7be8a7535a31e38820af80659758c

sneako (approved):

drewolson:
Just a quick check in that we've been using this in our QA environment for the past few days and it has worked great. Please let me know if there's more needed before merging @keathley.

keathley:
@drewolson Sounds good. I'm taking a look over it now. I suspect we'll be g2g but just want to confirm.

drewolson:
> @drewolson Sounds good. I'm taking a look over it now. I suspect we'll be g2g but just want to confirm.

Just checking back in after the weekend. Apologies if I'm being too persistent here, but I would love to get this merged.

keathley:
@drewolson Sorry for  the delay, this has been merged into main.

drewolson:
Thanks!

PR #93, opened 2020-10-23
Use lifo strategy to improve connection reuse

josevalim:


sneako (approved):
❤️  Thanks!!

For some background, I managed to throw together a script for José that made it easy to reproduce the connection: :closed errors that I am seeing consistently in prod.

He figured out that adding a LIFO strategy to NimblePool greatly improves connection reuse, since we are opening connections lazily.

Running my test script with these changes very significantly reduced the connection closed errors. These errors happen mostly for me when making requests against an Apache host, which does support Keep-Alive, but with a timeout of only 1 second...

keathley:
Thanks @josevalim!

PR #94, opened 2020-10-23
HTTP2 pool should enter disconnected state if there are no requests

keathley:
We've seen issues where h2 connections can get deadlocked in a read-only state. The real root cause seems to be that we're not able to reliably receive a socket closed message and it's unclear as to why. This change forces the gen_statem process into the disconnected state if there are no more pending requests. There is more bookkeeping with this approach, but it should help avoid the deadlock scenario.


drewolson:
We've also noticed that HTTP2 pools eagerly connect on boot. Should HTTP2 pools wait for an initial request before attempting to connect?

keathley:
I think we will still want to keep the pool alive in order to lower tail latencies. Because http2 can be multiplexed, it should be possible to get away with a very small number of connections which should mean that you don't have connections sitting there idle that often.

PR #95, opened 2020-10-24
Revert "Use lifo strategy to improve connection reuse"

sneako:
Reverts keathley/finch#93

These changes greatly improved things in the test scenario that I came up with to reliably reproduce the connection closed errors, but when I deployed them to production I actually saw the error rate increase 😦 .

We are continuing to investigate another strategy.


keathley (approved):

josevalim:
This should be closed in favor of #96 . :)

PR #96, opened 2020-10-24
Use lazy pools instead of lifo pools

josevalim:


sneako:
Amazing, thank you! I'll test this out in the morning and let you know how it goes

PR #98, opened 2020-10-25
Reduce runtime time unit conversions

josevalim:
Convert max_idle_time to native on init to
avoid doing conversions at runtime.

Also compute idle_time only once.

### DIFF
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -71,7 +71,7 @@ defmodule Finch.Conn do
     System.convert_time_unit(idle_time, :native, unit)
   end
 
-  def reusable?(%{max_idle_time: :infinity}, idle_time), do: true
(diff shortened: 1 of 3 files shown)


sneako (approved):

lib/finch/http1/conn.ex
```diff
@@ -71,11 +71,8 @@ defmodule Finch.Conn do
     System.convert_time_unit(idle_time, :native, unit)
   end
 
-  def reusable?(%{max_idle_time: :infinity}), do: true
-
-  def reusable?(conn) do
-    idle_time(conn, :millisecond) < conn.max_idle_time
-  end
+  def reusable?(%{max_idle_time: :infinity}, idle_time), do: true
```
sneako, line 74:
```suggestion
  def reusable?(%{max_idle_time: :infinity}, _idle_time), do: true
```

PR #102, opened 2020-11-07
Replace deprecated `:one_of`NimbleOption type

trarbr:
The `{:one_of, choices}` type was recently renamed to `{:in, choices}`: https://github.com/dashbitco/nimble_options/pull/54


sneako:
Thanks @trarbr! If we are going to make this change, then we should probably also update mix.exs too, making the requirement more strict, so that we do not pull in an older version of NimbleOptions, from before they introduced this change.

trarbr:
Good catch. I bumped the version in `mix.exs` as well.

sneako (approved):

PR #103, opened 2020-11-09
Update nimble_options/nimble_pool to eliminate deprecation message

begedin:
When adding fitch to a repository, since the version for nimble_options is set to 0.3, we get the latest revision, at the time of writing this, 0.3.5. That one supports the following two formats for `:protocol` in the pool config schema:

```Elixir
# new format
protocol: [
  type: {:in, [:http2, :http1]},
   # ...
]

# old format
protocol: [
  type: {:one_of, [:http2, :http1]},
   # ...
]
```

The old format is deprecated, so when I try to use finch in my application, I get a lot of console output:

```
warning: the {:one_of, choices} type is deprecated. Use {:in, choices} instead.
  (nimble_options 0.3.5) lib/nimble_options.ex:652: NimbleOptions.validate_type/1
  (nimble_options 0.3.5) lib/nimble_options.ex:509: NimbleOptions.validate_type/3
  (nimble_options 0.3.5) lib/nimble_options.ex:363: NimbleOptions.validate_option/3
  (nimble_options 0.3.5) lib/nimble_options.ex:343: NimbleOptions.reduce_options/2
  (elixir 1.10.4) lib/enum.ex:3686: Enumerable.List.reduce/3
  (elixir 1.10.4) lib/enum.ex:2161: Enum.reduce_while/3
  (nimble_options 0.3.5) lib/nimble_options.ex:336: NimbleOptions.validate_options/2
  (nimble_options 0.3.5) lib/nimble_options.ex:311: NimbleOptions.validate_options_with_schema_and_path/3
  (nimble_options 0.3.5) lib/nimble_options.ex:343: NimbleOptions.reduce_options/2
  (elixir 1.10.4) lib/enum.ex:3686: Enumerable.List.reduce/3
  (elixir 1.10.4) lib/enum.ex:2161: Enum.reduce_while/3
  (nimble_options 0.3.5) lib/nimble_options.ex:336: NimbleOptions.validate_options/2
  (nimble_options 0.3.5) lib/nimble_options.ex:311: NimbleOptions.validate_options_with_schema_and_path/3
  (nimble_options 0.3.5) lib/nimble_options.ex:228: NimbleOptions.validate/2
  (finch 0.5.1) lib/finch.ex:112: Finch.pool_options!/1
  (finch 0.5.1) lib/finch.ex:86: Finch.start_link/1
  (stdlib 3.13) supervisor.erl:385: :supervisor.do_start_child_i/3
  (stdlib 3.13) supervisor.erl:371: :supervisor.do_start_child/2
```

This happens once per finch pool being started i application, but it's troublesome in my case in tests, because I tend to have each test run it's own pool process, to keep things clearly separated, so I get this once per test.

0.3.0 does not support the `:in` version, so my options were to 

A. update the `:nimble_options` dependency. I updated `:nimble_pool` as well, since I assume the two are updated together, but I will admit I did not take the time to check.

B. lock the dependency to before this change was introduced, but from their change log, it is not clear which revision between 0.3.0 and 0.3.5 introduced it. If that's preferable, when I get the chance, I will check their commits and create that PR. It is my opinion, however, that in the long run, it's better to just update, which is what this PR does.


sneako:
Thanks @begedin! We are already looking at this here: https://github.com/keathley/finch/pull/102#pullrequestreview-526384049

sneako:
Also, if you want to get rid of the warnings right now, you could add `{:nimble_options, "0.3.0"}` to your mix.exs. Thanks again!

begedin:
Apologies, completely missed that. Thank you!

PR #109, opened 2020-12-16
Fix formatting

mrjoelkemp:
The current indentation makes the rendered documentation look like:

image

The fix will make the first section look like the others. 

---

I verified it using an online markdown renderer:

With the fix:

image

Without the fix: 

image


sneako (approved):
Thank you!

PR #114, opened 2021-02-02
Added an example to the start_link documentation

k-cross:
I didn't RTFM close enough and it was listed, but I think having an example would still be beneficial for non-default configuration options.

Code is all functionally the same. Close it if you feel like it's noise. Cheers!

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -46,8 +46,12 @@ defmodule Finch do
     ],
     conn_opts: [
       type: :keyword_list,
-      doc:
-        "These options are passed to `Mint.HTTP.connect/4` whenever a new connection is established. `:mode` is not configurable as Finch must control this setting. Typically these options are used to configure proxying, https settings, or connect timeouts.",
+      doc: """
+      These options are passed to `Mint.HTTP.connect/4` whenever a new connection is \
+      established. `:mode` is not configurable as Finch must control this setting. \
+      Typically these options are used to configure proxying, https settings, or connect \
+      timeouts.
+      """,
       default: []
     ]
   ]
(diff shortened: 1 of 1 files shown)


lib/finch.ex
```diff
@@ -80,6 +84,22 @@ defmodule Finch do
...
+    {
+      Finch,
+      name: MyFinch,
+      pools: %{
+        default: [
+          size: 10,
+          count: 1,
+          protocol: :http1,
+          max_idle_time: :infinity,
+          conn_opts: [transport_opts: [ciphers: :ssl.cipher_suits(:all)]]
```
wojtekmach, line 99: please don’t use :ssl.cipher_suits/1 as it’s going away in OTP 24 :)

k-cross: Updated to used `:ssl.cipuer_suits/2`

sneako:
I generated the docs from this PR and the example that is added felt a bit repetitive to me, since we already have a similar example in the usage section, see [here](https://hexdocs.pm/finch/Finch.html). Also, the spacing is a bit off so the code formatting is slightly broken, see screenshot.
Screen Shot 2021-02-16 at 09 23 28

 Do you thinking adding a few more configuration options to the existing example in the "Usage" section could be as effective?

k-cross:
@sneako yeah, that works too. I must have originally missed the example in the main page. It'd be great if there was a place to put the same example in both places but in a way that avoids duplication but that's probably overkill. I do think it's nice to have examples in both the functions that describe the behavior as well as top level since it is about general setup.

I'll make the change once I can get back to my personal computer.

sneako:
I'm going to close this due to inactivity, but please feel free to submit a new PR!

PR #116, opened 2021-02-11
Fix pool timeout test

nickjacob:
I was using this test to check what the full `{:error, _}` tuple is when the pool times out & found that the assert in the catch wasn't being called.

This doesn't affect functionality at all, just fixing this test so it does get an exit there. I based the changes off of [this nimble_pool test](https://github.com/dashbitco/nimble_pool/blob/master/test/nimble_pool_test.exs#L280)


sneako:
Great catch, thank you!

sneako (approved):

PR #126, opened 2021-02-27
Add helpful checkout timeout error

### LINKED ISSUE #125: Return helpful error on pool NimblePool.checkout timeouts
Right now the error returned from a checkout timeout in finch is very minimal and doesn't help a user solve the issue.

Capture and return a more helpful error to help the user know what options to tweak to resolve the issue.

SophisticaSean:
fixes #125

### DIFF
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -55,7 +55,23 @@ defmodule Finch.HTTP1.Pool do
     catch
       :exit, data ->
         Telemetry.exception(:queue, start_time, :exit, data, __STACKTRACE__, metadata)
-        exit(data)
+
+        # Provide helpful error messages for known errors
+        case data do
+          {:timeout, {NimblePool, :checkout, pids}} ->
+            reraise(
+              "
+            Timeout while waiting to checkout an http connection for process(es): #{inspect(pids)}.
+            Consider:
+            1. Increasing your poolsize (recommended).
+            2. Increasing your pool_timeout. Current timeout is: #{inspect(pool_timeout)}.
+            ",
+              __STACKTRACE__
+            )
+
+          _ ->
+            exit(data)
+        end
     end
   end
 


SophisticaSean:
I'm also running mix format on the files I've changed, so that's the source of any unrelated changes.

lib/finch/http1/pool.ex
```diff
@@ -55,7 +55,23 @@ defmodule Finch.HTTP1.Pool do
     catch
       :exit, data ->
         Telemetry.exception(:queue, start_time, :exit, data, __STACKTRACE__, metadata)
-        exit(data)
+
+        # Provide helpful error messages for known errors
+        case data do
```
SophisticaSean, line 60: switched from an `exit` to a `reraise` here to get the output to show up properly. If y'all know of a way to still use `exit` but get this nice error out then I'm happy to switch back to `exit`.

lib/finch/http1/pool.ex
```diff
@@ -55,7 +55,23 @@ defmodule Finch.HTTP1.Pool do
...
+          {:timeout, {NimblePool, :checkout, pids}} ->
+            reraise(
+              "
+            Timeout while waiting to checkout an http connection for process(es): #{inspect(pids)}.
+            Consider:
+            1. Increasing your poolsize (recommended).
```
keathley, line 66: I really like the intentionality behind this error message, but I don't know that we should make either of these recommendations. If the user is receiving this error, the obvious implication is that the connections are all being utilized longer than the request is willing to wait. This means we have a queue building up which further implies that our arrival rate is exceeding the work time of the downstream service.

Increasing the pool timeout is just going to allow that queue to grow even deeper making the problem worse by further taxing the systems resources.
Increasing the connection count per pool or adding a new pool _may_ help alleviate the issue. But, if the downstream system is becoming overloaded then adding additional load results in more queueing in the users service further degrading performance.

In any case, I don't feel like we should make either of these recommendations and instead should try to be verbose about _why_ they're seeing timeouts. Maybe something like:

```
Finch was unable to provide a connection within the timeout due to excess queueing for connections. This could be because the downstream system can't keep up with the amount of requests being sent or because you have an insufficient number of connections and need to increase your pool counts.
```

SophisticaSean: I'm all for it! I totally agree that the current message could lead a user down a poor path.

Any sort of language that doesn't lead a user to believe its a nimble pool problem at first glance is great IMO. As someone who has dealt with these cryptic-esque errors in hackney in production it can be hard to see the forest from the trees in the default error and understand exactly what you need to do to begin to ameliorate the issue.

I'll make the change to your verbiage.

SophisticaSean: change has been made verbatim.

sneako: > ```
> Finch was unable to provide a connection within the timeout due to excess queueing for connections. This could be because the downstream system can't keep up with the amount of requests being sent or because you have an insufficient number of connections and need to increase your pool counts.
> ```

In my experience the pool `count` should be increased if you see the pool becoming a bottleneck (message queue pile up) or also in some cases when you you see the [queue duration](https://github.com/keathley/finch/blob/main/lib/finch/telemetry.ex#L26) increase. In the case of checkout timeouts, you usually want to try increasing the `size` first.

lib/finch/http1/pool.ex
```diff
@@ -55,7 +55,22 @@ defmodule Finch.HTTP1.Pool do
     catch
       :exit, data ->
         Telemetry.exception(:queue, start_time, :exit, data, __STACKTRACE__, metadata)
-        exit(data)
+
+        # Provide helpful error messages for known errors
+        case data do
+          {:timeout, {NimblePool, :checkout, _affected_pids}} ->
+            reraise(
```
sneako, line 62: How about something like this?
```elixir
    reraise(
      """
      Finch was unable to provide a connection within the timeout. Consider adjusting \
      the pool size, count, timeout or reducing the rate of requests if it is possible \
      that the downstream service is unable to keep up with the current rate.
      """,
      __STACKTRACE__
    )
```
Besides the updated message, also note the usage of triple quote and backslashes to avoid awkward whitespace

test/finch_test.exs
```diff
@@ -392,9 +398,9 @@ defmodule FinchTest do
 
       try do
         Finch.build(:get, endpoint(bypass)) |> Finch.request(client, pool_timeout: 0)
-      catch
-        :exit, reason ->
-          assert {:timeout, _} = reason
+      rescue
+        e in RuntimeError ->
+          assert e.message =~ "Finch was unable to provide a connection within the timeout"
```
wojtekmach, line 403: you may consider converting it to a more common:

```elixir
assert_raise RuntimeError, ~r/Finch was unable to provide a connection within the timeout/ fn ->
  # ...
end
```

That being said, and this is out of the scope of this PR, but curious if instead of raising an error have you considered returning an `{:error, %RuntimeError{}}` tuple instead?

sneako: I like the idea of changing this to an error tuple instead of raising. I think this only raises because NimblePool does, so we kept it the same here

josevalim: I would prefer to raise instead of a tuple because I can't think of a scenario where you want to handle this. Even if you say, "well, I want to retry", that's bound to make things worse. :)

sneako: That definitely makes sense

mattbaker:
We're debugging some nimble pool checkout issues and I found this PR, it was educational for me and might be a good addition to Finch!

daveshah:
Our team was looking into some nimble pool checkout issues as well today. 
💯 % agree with @mattbaker  - this PR was helpful. Given that this is over 1 year old, have any other changes been made to make errors in this area a bit more clear? If not, would it make sense to resolve the conflicts here and try to get this in?

sneako:
Alright, I'm convinced! Glad that it has proven to be useful. I will clean things up and merge it soon.

SophisticaSean:
Updated the PR with upstream so we can get this merged!

test/finch_test.exs
SophisticaSean: fixed!

daveshah:
Thanks @SophisticaSean! 🙌🏽

lib/finch/http1/pool.ex
```diff
@@ -54,7 +54,22 @@ defmodule Finch.HTTP1.Pool do
...
+              """
+              Finch was unable to provide a connection within the timeout due to excess queueing for connections.\
+              This could be because the downstream system can't keep up with the amount of requests being sent or\
+              because you have an insufficient number of connections and need to increase your pool counts.
+              """,
```
sneako, line 66:
```suggestion
              """
              Finch was unable to provide a connection within the timeout due to excess queuing \
              for connections. Consider adjusting the pool size, count, timeout or reducing the \
              rate of requests if it is possible that the downstream service is unable to keep up \
              with the current rate.
              """,
```

PR #138, opened 2021-05-10
Update CI config

wojtekmach:
- Remove Elixir 1.10 from buildl matrix
- Update actions/cache usage (see https://github.com/actions/cache/pull/469#issuecomment-823346224)


sneako:
@wojtekmach any idea why the lint conditionals I added don't seem to be skipping those steps? 🤔

PR #145, opened 2021-07-14
Stop appending empty query to path

mruoss:
If I pass a request object with `query: ""`, I end up with a needless "?" appended to the path. I think `request_path/1` should also check for query being empty string. What do you think?


sneako (approved):
Thanks @mruoss! LGTM

Issue #150, opened 2021-09-09
HTTP monitor leak


NeilMenne:
While charting memory utilization in a service at B/R, I noticed that Finch was prominently in the top spot in the `:observer_cli`, so I confirmed my hypothesis that the memory was increasing overall.

Looking at, `:recon.info(pid) |> get_in([:signals, :monitored_by]) |> length()` we see that the monitors are increasing seemingly per-request. Since my requests are being made my `GenServer`s maintaining a cache, the processes they get linked to are long-lived.

@keathley helped me confirm diagnosis against the Finch code, so he should be able to help resolve this in a timely manner.

keathley:
This should now be resolved with 0.8.2. Re-open if you continue to see issues.

PR #151, opened 2021-09-09
Demonitor process when request is done

keathley:
We don't demonitor the http connection when making http/2 requests. We need to do this in order to avoid memory leaks and build up of useless monitors.


keathley:
This fixes #150

Issue #139, opened 2021-05-17
Support unix sockets


dallagi:
Hello,

Mint natively supports making requests on unix sockets by setting the address to `{:local, "/path/to/socket.sock"}` and the port number to `0`, for example:
```elixir
{:ok, conn} = Mint.HTTP.connect(:http, {:local, "/var/run/docker.sock"}, 0, hostname: "localhost")
```
For more context, tests for this functionality in Mint can be found [here](https://github.com/elixir-mint/mint/blob/main/test/mint/unix_socket_test.exs).

However, to my knowledge Finch does not currently allow this.
In case this feature may be desirable, a possible approach could be to do as Hackney does, ie. [supporting url schemas in the form of eg. `"http+unix"`](https://github.com/benoitc/hackney/blob/6e79b2bb11a77389d3ba9ff3a0828a45796fe7a8/src/hackney_url.erl#L49).

sneako:
Hi and thanks for the issue! I agree that Finch should support this. I have not looked in to this very much yet, but rather than using a scheme like `http+unix`, I think we should use the `{:local, socket_path}` tuple that Mint uses, which is the type from inet https://erlang.org/doc/man/inet.html#type-local_address.

dallagi:
Hi @sneako , thanks for your quick reply!

Personally I don't have a strong opinion about which approach to use, but I'm afraid that using the `local_address()` type from inet would not fit well with the current API exposed by Finch, which (if I understand correctly) currently expects a `URI` or a string that can be parsed to a `URI`.

Anyway, once the preferred approach is defined I would be happy to implement it and send a PR, if that's OK for you :)

sneako:
Hi @dallagi, sorry for the delay in getting back to you. I have been thinking about this a bit in the meantime, and have come to the conclusion that my initial proposal of using the local_address type would probably not be sufficient. I have not personally used TLS over a unix socket, however based on my limited research it does seem to be entirely possible. There is no way to specify that you would like to use TLS in the local_address tuple that I had proposed, therefore your proposal of `http+unix` or `https+unix` does seem like it could be the best way forward.

dallagi:
Hi @sneako,
No worries at all for the delay, I appreciate the effort you’re putting in maintaining finch :)

Thanks for the heads up about the potential issues!
I agree that having to url encode the path of the socket is not very convenient.

However, personally I can’t think of a better alternative.
[...]

PR #153, opened 2021-10-13
Add support to unix sockets

cgerling:
Closes https://github.com/keathley/finch/issues/139

This PR doesn't add support to connect with UNIX sockets through HTTPS, I wasn't sure how to do it without changing a lot of stuff so let me know if there is a "right way" to implement it


cgerling:
This PR doesn't add support to connect with UNIX sockets through HTTPS, I wasn't sure how to do it without changing a lot of stuff so let me know if there is a "right way" to implement it

sneako:
Great work @cgerling ! Just a couple minor comments, but overall this looks good!

PR #154, opened 2021-10-14
Bump up CI min OTP version to allow running on latest OTP version

sneako:
Due to the changes that allowed our test suite to run on OTP 24,
we need to stop running it on OTP 19.

Also updating the other version pair to the latest releases.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -1,5 +1,9 @@
 name: CI
-on: [pull_request, push]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
 jobs:
  mix_test:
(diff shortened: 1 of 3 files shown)


sneako:
I was hoping to continue running CI on OTP 20, but if we want the test suite to run on OTP 24 we need the latest version of Cowboy which only supports OTP 22+. Locally I am running the lastest OTP & Elixir and I assume most contributors will be running OTP 24 locally as well, so I think it is more important to fully support this version.

We can look in to replacing Cowboy with something else if we really want to have a wider test matrix, but at least our main deps, Mint & NimblePool, both run CI on older OTP versions.

PR #160, opened 2021-11-03
Add request transformer callback, tests, and module docs

learc83:
While trying to add OpenTelemetry tracing to some of our applications, we couldn't find an easy solution to inject tracing headers into all requests. We'd love to hear about it if a solution already exists, but this PR is our proposed solution if not.

If this PR is accepted, we'll be able to create an OpenTelemetryFinch library similar to OpenTelemetryPhoenix.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -79,18 +79,30 @@ defmodule Finch do
     * `:name` - The name of your Finch instance. This field is required.
 
     * `:pools` - A map specifying the configuration for your pools. The keys should be URLs
-    provided as binaries, a tuple `{scheme, {:local, unix_socket}}` where `unix_socket` is the path for 
+    provided as binaries, a tuple `{scheme, {:local, unix_socket}}` where `unix_socket` is the path for
     the socket, or the atom `:default` to provide a catch-all configuration to be used for any
-    unspecified URLs. See "Pool Configuration Options" below for details on the possible map 
+    unspecified URLs. See "Pool Configuration Options" below for details on the possible map
     values. Default value is `%{default: [size: #{@default_pool_size}, count: #{@default_pool_count}]}`.
 
+    * `:request_transformer` - A callback that can be used to modify Requests as they're being
+    made (for example, to inject distributed tracing headers). The arity-1 function will receive
+    the `t:Finch.Request.t/0` and must also return the transformed `t:Finch.Request.t/0`. Note that
+    this function will be called synchronously during every request, so care should be taken to ensure
+    that it does not introduce unnecessary latency.
   ### Pool Configuration Options
 
   #{NimbleOptions.docs(@pool_config_schema)}
+
+  ### Request Transformer Example
+
+      request_transformer: fn req ->
+        %{req | headers: [{"injected-header", "123"} | req.headers]}
+      end
   """
   def start_link(opts) do
     name = Keyword.get(opts, :name) || raise ArgumentError, "must supply a name"
     pools = Keyword.get(opts, :pools, []) |> pool_options!()
+    request_transformer = Keyword.get(opts, :request_transformer, nil)
     {default_pool_config, pools} = Map.pop(pools, :default)
 
     config = %{
@@ -98,7 +110,8 @@ defmodule Finch do
       manager_name: manager_name(name),
       supervisor_name: pool_supervisor_name(name),
       default_pool_config: default_pool_config,
-      pools: pools
+      pools: pools,
+      request_transformer: request_transformer
     }
 
     Supervisor.start_link(__MODULE__, config, name: supervisor_name(name))
@@ -231,6 +244,9 @@ defmodule Finch do
           {:ok, acc} | {:error, Exception.t()}
         when acc: term()
   def stream(%Request{} = req, name, acc, fun, opts \\ []) when is_function(fun, 2) do
+    {:ok, config} = Registry.meta(name, :config)
+    req = if config.request_transformer, do: config.request_transformer.(req), else: req
+
     shp = build_shp(req)
     {pool, pool_mod} = PoolManager.get_pool(name, shp)
     pool_mod.request(pool, req, acc, fun, opts)
--- a/test/finch_test.exs
+++ b/test/finch_test.exs
@@ -786,6 +786,28 @@ defmodule FinchTest do
     end
   end
 
+  describe "request-transformer" do
+    test "adds headers to request", %{bypass: bypass} do
+      start_supervised!(
+        {Finch,
+         name: MyFinch,
+         request_transformer: fn req ->
+           %{req | headers: [{"injected-header", "123"} | req.headers]}
+         end}
+      )
+
+      Bypass.expect_once(bypass, "GET", "/", fn conn ->
+        assert Enum.member?(conn.req_headers, {"injected-header", "123"})
+        assert Enum.member?(conn.req_headers, {"build-header", "abc"})
+        Plug.Conn.send_resp(conn, 200, "OK")
+      end)
+
+      assert {:ok, %{status: 200}} =
+               Finch.build(:get, endpoint(bypass), [{"build-header", "abc"}])
+               |> Finch.request(MyFinch)
+    end
+  end
+
   defp get_pools(name, shp) do
     Registry.lookup(name, shp)
   end


sneako:
Hi @learc83! Thanks for looking in to this, I would love to see an OpenTelemetryFinch library.

My initial concern was that adding this `Registry.meta/2` call for every request would have some performance impact, however I ran some tests and the impact is negligible :partying_face:  

I'd like to leave this open for a few days to allow for community feedback, but IMO this looks good!

learc83:
Awesome! Sounds good to me.

GregMefford:
One concern I have is that there is only one callback function rather than allowing multiple things to register themselves, so if you wanted to have a list, you'd need a custom function that chains them together. Maybe that's fine because the order actually does matter when you chain them together, but thought it was worth raising in case there is another related concern there.

sneako:
Hi @GregMefford! If I am understanding correctly, you are referring to a case where someone wants to use multiple request_transformer functions? Since this function accepts a Request and returns a Request, they should compose quite nicely, very similar to the Plug interface, and this composition could happen in the function that is passed in to the `:request_transformer`.

Speaking of Plug, perhaps we should follow that design even more closely by making the request transformer function accept 2 args, the request and an optional keyword list of options, which could be the opts that are passed to `stream/5` or `request/3`. This would allow the request transformer to behave differently for different requests if desired.

learc83:
> Speaking of Plug, perhaps we should follow that design even more closely by making the request transformer function accept 2 args, the request and an optional keyword list of options, which could be the opts that are passed to `stream/5` or `request/3`. This would allow the request transformer to behave differently for different requests if desired.

I like that idea. I've added it in the latest commit, along with an example test showing how you could use it to skip injecting headers.

GregMefford:
> Hi @GregMefford! If I am understanding correctly, you are referring to a case where someone wants to use multiple request_transformer functions? Since this function accepts a Request and returns a Request, they should compose quite nicely, very similar to the Plug interface, and this composition could happen in the function that is passed in to the `:request_transformer`.

Yes, that was what I was thinking, more specifically in the case where we had another library that tells people to point this transformer function directly at that library (e.g. to inject OTel tracing headers), then it might not be obvious to them how they would make a custom function to chain multiple transformers together. I already convinced myself that isn’t really a concern though, and we could just document that in the other library’s integration docs - that’s not really Finch’s concern unless there was a desire for more complicated middleware pipelines or something (which I don’t think we want for now at least).

sneako:
Maybe we should also formally specify the RequestTransformer behaviour with a callback module? WDYT?

learc83:
> Maybe we should also formally specify the RequestTransformer behaviour with a callback module? WDYT?

Sounds like a nice ergonomic improvement. I've added this.

lib/finch.ex
```diff
@@ -231,6 +236,13 @@ defmodule Finch do
           {:ok, acc} | {:error, Exception.t()}
         when acc: term()
   def stream(%Request{} = req, name, acc, fun, opts \\ []) when is_function(fun, 2) do
+    {:ok, config} = Registry.meta(name, :config)
+
+    req =
+      if config.request_transformer,
+        do: config.request_transformer.transform(req, opts),
```
GregMefford, line 243: Not to belabor this PR any further, but I have been asking myself if there's also value in passing `name` through as well, or is that already something you could infer from the `Request` struct?

lib/finch/request_transformer.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Finch.RequestTransformer do
+  @moduledoc """
+  A RequestTransformer behaviour module that specifies a `transform/2` callback that can be used to modify
+  Requests as they're being made (for example, to inject distributed tracing headers). Opts is passed
+  down from the opts given to `stream/5` or `request/3`, so it may be used to allow the callback to behave
+  differently per request. Note that this function will be called synchronously during every request, so care
+  should be taken to ensure that it does not introduce unnecessary latency.
```
GregMefford, line 7: Just a suggestion (I didn't word-wrap because I'm writing this in GitHub, so ignore whitespace in my suggestion 😅):
```suggestion
    A behaviour for transforming a `Finch.Request` struct just before the actual request is made, for example, to inject distributed tracing headers.
  
  The `transform/2` callback will be called with the following parameters and should return the transformed `Finch.Request.t()` struct:
  
   * `request`: The `Finch.Request.t()` struct representing the request to be transformed.
   * `opts`: the options that were passed to `Finch.stream/5` or `Finch.request/3`, so that the transformer can behave differently per request.
   
   Note that this function will be called synchronously during every request, so care should be taken to ensure that it does not introduce unnecessary latency.
```

lib/finch.ex
learc83: I don't think there's a way to infer name, so if you think it would be useful now would be the time to add it. Any specific use cases you can think of?

GregMefford: I was just thinking maybe you transform differently somehow based on which Finch it is, e.g. what downstream service you’re calling. Now that I think about it though... maybe you'd never need to really do that because you could specify a different `request_transformer` for each Finch supervisor at setup time if you wanted to do that, rather than pointing them all at the same function that dispatches based on name. It's really the same thing I think. 🤷

learc83: It's easy enough to add (and to ignore if you don't need it) that I went ahead and did it. Open to feedback on whether to keep it there or not.

sneako:
The more I think about this, I don't see any real advantage to adding this option versus wrapping your Finch calls in a module that injects the tracing headers.

If we add this option, every pool for every instance of finch would need to specify this new config option.

Comparing that to a module, maybe even the OpenTelemetryFinch module itself, that provides a wrapped build and/or request function to inject these headers or perform any other kinds of request transformations, doesn't seem as clean.

I know I said I did not observe any performance impact in the bit of testing that I did, but at the end of the day, this change forces Finch to do a bit more work for every single request, even for users that are not interested in using the request transformer.

What do you think?

Is there any real advantage to adding this versus wrapping the Finch calls?

GregMefford:
For me, the main disadvantage to wrapping Finch is that then the integration library (like a hypothetical OTelFinch that adds tracing headers) needs to know about and keep up with the entire Finch API (at the bare minimum, having a `defdelegate` for each function head), and it's up to the that library author to make sure they're keeping up with any new features in Finch that might make the integration not work for some reason in new versions of Finch, even though they're not technically breaking changes from Finch's perspective (e.g. some new top-level function to make requests a different way or with different options). With this transformation API, the integration library would just need to make sure that they are fulfilling the defined behaviour contract, and Finch would be responsible to make sure that behaviour is invoked in the right places within the Finch codebase if new features are added.

Regarding performance, I'm not very familiar with the Finch internals, but I was surprised that the config would even need to be fetched from Registry synchronously during each request - if there were some way to have the config be more efficiently passed to this function (or even just the transformer specifically since that's what we need here), then that could help to mitigate any performance concern. For example, if you could store the transformer callback in `persistent_term` based on the `name`, then you could avoid the Registry lookup and I think it would be faster. Maybe Finch isn't ready to rely on `persistent_term` being available yet from a compatibility standpoint, but it's something to consider as an option for the future if not now.

sneako:
Thanks for clarifying @GregMefford! I do see your point, and I feel that the burden of keeping up with the Finch API could be minimized. I have been considering stabilizing Finch's public API, meaning removing the deprecated `Finch.request/6` and publishing a 1.0.0 release.

In that case, a hypothetical OpenTelemetryFinch module could look something like:

```elixir
defmodule OpenTelemetryFinch do
  defdelegate start_link(opts), to: Finch
  defdelegate child_spec(init_arg), to: Finch
  defdelegate build(method, url, headers \\ [], body \\ nil, opts \\ []), to: Finch

  def request(req, name, opts \\ []) do
    req
    |> inject_tracing_headers(opts)
    |> Finch.request(name, name, opts)
  end

  def stream(req, name, acc, fun, opts \\ []) do
    req
    |> inject_tracing_headers(opts)
    |> Finch.stream(name, acc, fun, opts)
  end

  def inject_tracing_headers(opts) do
    # ...
  end
end
```

Would that alleviate your concerns?

GregMefford:
Yeah in practice it isn’t so bad, it just kind of feels weird to me ergonomically to wrap APIs like that. I know some vendors have wrapper libraries like that for the Task module, for example, and I always grumble when they make me use their wrapper because it makes things harder to extend or compose. Like if I wanted to use two wrappers at the same time, I can’t because they both directly use the “real” base library so you typically can’t tell one to use the other or whatever.

If you don’t want to support a transformer in Finch, though, this is a viable work-around.

josevalim:
A global setting sounds dangerous. Someone can use it for instrumentation but then someone can accidentally leak authentication headers to other services too. Wouldn't it suit better per pool?

keathley:
Just my two cents, but if transformers are going to be allowed, it seems useful to take a similar approach to what @wojtekmach has done in req. That way everything can be arbitrarily composed together. Alternatively, stuff like otel or tracing headers or whatnot could target req or Tesla middleware.

GregMefford:
> A global setting sounds dangerous. Someone can use it for instrumentation but then someone can accidentally leak authentication headers to other services too. Wouldn't it suit better per pool?

That’s another good reason to have this be a configuration setting on each Finch instance rather than a wrapper library, where it would be harder to introspect / configure whether to transform things or not. I suppose you could just decide at each call site whether to call OTelFinch or just regular Finch to indicate whether headers should be injected, though.

GregMefford:
> Just my two cents, but if transformers are going to be allowed, it seems useful to take a similar approach to what @wojtekmach has done in req. That way everything can be arbitrarily composed together. Alternatively, stuff like otel or tracing headers or whatnot could target req or Tesla middleware.

Yeah, I guess a practical solution to the OTel problem could just be to tell people to use Req or Tesla + OTel middleware instead of using Finch directly, if the Finch maintainers don’t want to add the complexity of this “middleware” to Finch itself. That’s totally fair if that’s where we end up. This PR was mostly just to open that discussion with a minimal-change implementation to the existing APIs.

keathley:
I’m not saying they shouldn’t be added really. I’m also not advocating for them either. That’s @sneako’s call. But, if it’s gunna be added, it seems like it should be arbitrarily extensible and req seems like a good model for that.

sneako:
Thanks for chiming in everyone!

The point about the wrapper module suggestion turning this into essentially a global setting is definitely valid and turns me off of that idea.

My concern here is not really related to complexity, but really more about the (however so minor) performance hit of doing another Registry call on every single request even for users who have no interesting in request transformers.

I did realize that we could actually enable this feature without the additional Registry call, since we are already querying the registry on every request (to get the pool pid and pool module). We could theoretically add the request transformers to the data that we are storing in the registry [here](https://github.com/sneako/finch/blob/main/lib/finch/http1/pool.ex#L66) and [here](https://github.com/sneako/finch/blob/main/lib/finch/http2/pool.ex#L128).

However, I am also concerned that if we add request transformers, then we would also need response transformers too and I don't think either of these features really align with Finch's main goal of performance. The way I see Finch is essentially a library that tries to provide good pooling strategies for Mint, but it certainly will never be a full featured and super ergonomic HTTP client.

Maybe it does make the most sense for Finch's OpenTelemetry integration to happen through Req and Tesla instead, both of which I do see as full featured, ergonomic clients and still allow users to enjoy our pooling strategies.

learc83:
@sneako 

How do you feel about making this into a less general solution? An `inject_headers` callback instead of a `request_transformer`.

I'd really like to see if we can come up with some way for people to easily instrument Finch without forcing them to move up the abstraction chain to another higher level library. I think it would help OTel adoption if instrumenting Finch was just a matter of including an otel_finch library and passing in a small amount of extra config information vs. forcing people to swap out their current HTTP client library everywhere they're using it.

sneako:
No apology necessary @learc83, and congrats on your new baby! I am a new father myself (coming up on 1 year, so I'm starting to get the hang of it), so I have a good idea of what you are going through :)

I'm not sure that we should limit the capabilities of this option, just so that I will be protected from future requests to add a response transformer, but I appreciate it if that was the intention.

I put together a quick [proof of concept](https://github.com/sneako/finch/compare/req-transformer-take-2?expand=1) to demonstrate how we could potentially store and retrieve the request transformer without an additional registry lookup. 

Have you thought a bit more about what else the hypothetical OpenTelemetryFinch might need? Would this feature plus the existing Telemetry events be enough?

For example, in the case where Finch in running in an app that receives a request that already has a trace id header, I assume that OpenTelemetryFinch would want to propagate that same trace id, rather than generating a new one. Have you thought through how that might work?

sneako:
Closing due to inactivity

PR #165, opened 2021-12-24
add failing test with iolist > window

ruslandoga:
Continuing from https://github.com/sneako/finch/pull/158#commitcomment-62452183

PR #166, opened 2021-12-24
fix handling of iodata in HTTP/2 request streams

the-mikedavis:
depends #165

### DIFF
--- a/lib/finch/http2/request_stream.ex
+++ b/lib/finch/http2/request_stream.ex
@@ -6,19 +6,10 @@ defmodule Finch.HTTP2.RequestStream do
   def new(body, {from_pid, _from_ref} = from) do
     enumerable =
       case body do
-        {:stream, stream} -> stream
-        data -> List.wrap(data)
+        {:stream, stream} -> Stream.map(stream, &with_byte_size/1)
+        nil -> [with_byte_size("")]
+        io_data -> [with_byte_size(io_data)]
       end
-
-    enumerable =
-      Stream.map(enumerable, fn
-        binary when is_binary(binary) ->
-          {binary, byte_size(binary)}
-
-        io_data ->
-          binary = IO.iodata_to_binary(io_data)
-          {binary, byte_size(binary)}
-      end)
 
     reducer = &reduce_with_suspend/2
 
@@ -31,6 +22,9 @@ defmodule Finch.HTTP2.RequestStream do
       continuation: &Enumerable.reduce(enumerable, &1, reducer)
     }
   end
+
+  defp with_byte_size(binary) when is_binary(binary), do: {binary, byte_size(binary)}
+  defp with_byte_size(io_data), do: io_data |> IO.iodata_to_binary() |> with_byte_size()
 
   defp reduce_with_suspend(
          {message, message_size},
--- a/test/finch_test.exs
+++ b/test/finch_test.exs
@@ -379,14 +379,15 @@ defmodule FinchTest do
 
       # 2KB of data
       req_body = List.duplicate(125, 2_000)
-      response_body = req_body
+      req_body_binary = IO.iodata_to_binary(req_body)
+      response_body = req_body_binary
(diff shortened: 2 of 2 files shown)


test/finch_test.exs
```diff
@@ -379,14 +379,15 @@ defmodule FinchTest do
 
       # 2KB of data
       req_body = List.duplicate(125, 2_000)
```
ruslandoga, line 381: @the-mikedavis I wonder if it needs to be increased to 2MB as in the binary version of this test?

the-mikedavis: so I think this bug is actually more about mis-handling iodata than the iodata exceeding the window size (hence why you hit this with only 2KB when the default window is ~64KB if I remember correctly)

it probably makes sense to increase it to 2MB as you say though because that should effectively test both that iodata is handled correctly, and that window sizes are respected when using iodata

test/finch_test.exs
```diff
@@ -358,6 +358,58 @@ defmodule FinchTest do
...
+               transport_opts: [
+                 verify: :verify_none
+               ]
+             ]
+           ]
+         }}
+      )
+
+      # 2KB of data
+      req_body = List.duplicate(125, 2_000)
```
the-mikedavis, line 381:
```suggestion
      # 2MB of data
      req_body = List.duplicate(125, 2_000_000)
```

sneako:
Thanks @ruslandoga and @the-mikedavis!

lib/finch/http2/request_stream.ex
```diff
@@ -32,6 +23,9 @@ defmodule Finch.HTTP2.RequestStream do
     }
   end
 
+  defp with_byte_size(binary) when is_binary(binary), do: {binary, byte_size(binary)}
+  defp with_byte_size(io_data), do: io_data |> IO.iodata_to_binary() |> with_byte_size()
```
wojtekmach, line 27: if it's possible to leave this as a binary (and thus avoid copying) and use `IO.iodata_length/1` (still O(n) though) it might be a bit more efficient but I didn't benchmark this.

sneako: Good point! Thanks @wojtekmach, I’ll try it out

test/finch_test.exs
```diff
@@ -358,6 +358,58 @@ defmodule FinchTest do
...
+                 req_body
+               )
+               |> Finch.request(finch_name())
+
+      assert {_, ^header_val} =
+               Enum.find(headers, fn
+                 {^header_key, _} -> true
+                 _ -> false
+               end)
+    end
```
wojtekmach, line 411: should this be:

```elixir
assert {header_key, header_val} in headers
```

?

lib/finch/http2/request_stream.ex
sneako: I assume you meant "if it is possible to leave this as **iodata**". I tried quickly changing this to: `defp with_byte_size(io_data), do: {io_data, IO.iodata_length(io_data)}` but we are currently relying on binary pattern matching to extract binaries that will fit within the current window, therefore we do need to cast the iodata to a binary with the current implementation. I don't believe that there is a straightforward way to slice N bytes from iodata, is there?

wojtekmach: Yeah I meant to leave it as iodata and yeah, don’t think there is a way to efficiently slice it. :(

ruslandoga: Maybe something like https://github.com/ninenines/cowlib/blob/master/src/cow_iolists.erl could be used?

sneako: Nice find! We can try doing some benchmarking to see if this is worth it. I wonder if there was ever any discussion around adding this functionality to the stdlib...

test/finch_test.exs
sneako: https://github.com/sneako/finch/pull/167

PR #169, opened 2021-12-28
Finch.Request struct typespec

### LINKED ISSUE #168: Finch.Request.t() typespec
While looking at the hexdoc to get the type of some request fields, I've seen that the `Finch.Request.t()` struct is not "typespec-ed", or better it's defined as:

https://github.com/sneako/finch/blob/a5a5e3c740e4d14729d64e559f3a1a2047d1ec1d/lib/finch/request.ex#L52

This results to:
https://hexdocs.pm/finch/Finch.Request.html#t:t/0

The `@type` could be defined as:

```elixir
  @type t :: %__MODULE__{
          scheme: Mint.Types.scheme(),
          host: String.t(),
          port: :inet.port_number(),
          method: String.t(),
          path: String.t(),
          headers: headers(),
          body: body(),
          query: String.t(),
          unix_socket: nil | String.t()
        }
```

Is that intentional?
If not I can open a PR, just ping me here.

Thank you

visciang:
Fix #168

### DIFF
--- a/lib/finch/request.ex
+++ b/lib/finch/request.ex
@@ -49,7 +49,17 @@ defmodule Finch.Request do
   """
   @type body() :: iodata() | {:stream, Enumerable.t()} | nil
 
-  @type t :: %Finch.Request{}
+  @type t :: %__MODULE__{
+          scheme: Mint.Types.scheme(),
+          host: String.t(),
+          port: :inet.port_number(),
+          method: String.t(),
+          path: String.t(),
+          headers: headers(),
+          body: body(),
+          query: String.t(),
+          unix_socket: nil | String.t()
+        }
 
   @doc false
   def request_path(%{path: path, query: nil}), do: path
--- a/lib/finch/request.ex
+++ b/lib/finch/request.ex
@@ -51,14 +51,14 @@ defmodule Finch.Request do
 
   @type t :: %__MODULE__{
           scheme: Mint.Types.scheme(),
-          host: String.t(),
+          host: String.t() | nil,
           port: :inet.port_number(),
           method: String.t(),
           path: String.t(),
           headers: headers(),
           body: body(),
-          query: String.t(),
-          unix_socket: nil | String.t()
+          query: String.t() | nil,
+          unix_socket: String.t() | nil
         }
 
   @doc false


lib/finch/request.ex
```diff
@@ -49,7 +49,17 @@ defmodule Finch.Request do
...
-  @type t :: %Finch.Request{}
+  @type t :: %__MODULE__{
+          scheme: Mint.Types.scheme(),
+          host: String.t(),
+          port: :inet.port_number(),
+          method: String.t(),
+          path: String.t(),
+          headers: headers(),
+          body: body(),
+          query: String.t(),
```
sneako, line 60:
```suggestion
          query: String.t() | nil,
```

lib/finch/request.ex
```diff
@@ -49,7 +49,17 @@ defmodule Finch.Request do
...
+          unix_socket: nil | String.t()
```
sneako, line 61:
```suggestion
          unix_socket: String.t() | nil
```

sneako (changes requested):

lib/finch/request.ex
visciang: Actually the "nullable" fields are:
- host (from `URI.parse()`)
- port (from `URI.parse()`)
- body (from API contract)
- query (from `URI.parse()`)
- unix_socket (from API contract)

Maybe `Finch.Request.parse_url/1` should validate also `host` (to be not `nil`).

sneako: Can you provide an example `Finch.build/5` call that will result in `nil` for host? It does not appear to be possible since we require a scheme to be present.

visciang:
```!elixir
Finch.Request.build(:get, "http:", [], nil, [])          
%Finch.Request{
  body: nil,
  headers: [],
  host: nil,
  method: "GET",
  path: "/",
  port: 80,
  query: nil,
  scheme: :http,
  unix_socket: nil
}
```

sneako: Ah of course! I had not tried without `//`. In this case, I suppose it would make sense to raise an error if the host were parsed as `nil`, just like we raise when the scheme is invalid/unsupported.

wojtekmach: I believe the host is ignored when using the unix domain socket so maybe it is fine to leave this as is, and, if anything, let Mint raise if appropriate?

sneako: Good point, thanks @wojtekmach ! Let's leave it as is.

sneako (approved):
Thanks @visciang !

PR #171, opened 2022-02-19
implement pool idle timeout to http1 pools

oliveigah:
Addressing the issue https://github.com/sneako/finch/issues/159 this PR makes usage of the nimble pool feature terminate idle resources to implement idle pools termination to finch http1 pools.

Some considerations:

- The restart option for pools that use this feature needed to change, otherwise it keeps restarting over and over.
- IMO since it is possible to have the same shp tuple on multiple finch instances, would be nice to have the finch instance name on the telemetry event, but I could not find a proper way to do it.
- Might be nice to add some similar feature to http2 pools

Let me know what you think! :smile:


lib/finch.ex
```diff
@@ -57,6 +57,17 @@ defmodule Finch do
       used to configure proxying, https settings, or connect timeouts.
       """,
       default: []
+    ],
+    pool_max_idle_time: [
```
sneako, line 61: I think this name makes sense, but adding it makes the existing `:max_idle_time` option seem a bit vague. We should probably deprecate that option in favor of a new `:conn_max_idle_time`. What do you think?

README.md
```diff
@@ -54,8 +54,8 @@ children = [
 
 Pools will be started for each configured `{scheme, host, port}` when Finch is started.
 For any unconfigured `{scheme, host, port}`, the pool will be started the first time
-it is requested. Note pools are not automatically terminated if they are unused, so
-Finch is best suited when you are requesting a known list of static hosts.
+it is requested. Note pools are not automatically terminated by default, if you need to
+terminate them after some idle time, use the `pool_max_idle_time` option.
```
sneako, line 58: We should probably mention that this feature is only supported by HTTP1 pools for now.

sneako:
Just two small comments but overall, looks great!

oliveigah:
Sorry for the delay, rough week at work.

Also rough week on the world, hope you are well and safe.

I agree with both suggestions, gonna work on them.

sneako:
No apologies necessary! These are indeed rough times. There is no rush at all, please take as much time as you need :heart:

oliveigah:
Hey @sneako, I've add some modifications regarding the deprecation of the `max_idle_timeout` option (better late than never :sweat_smile: ).

Let me know if the change on the interface is enough or if you want me to go further and rename variables and add a telemetry event with the new name. I think it would be a nice touch to keep coherence among the interface, the code and the telemetry events but it is your call.

lib/finch.ex
```diff
@@ -47,7 +47,7 @@ defmodule Finch do
       The maximum number of milliseconds an HTTP1 connection is allowed to be idle \
       before being closed during a checkout attempt.
       """,
-      default: :infinity
+      deprecated: "This options is deprecated, use conn_max_idle_time instead"
```
sneako, line 50:
```suggestion
      deprecated: "This option is deprecated, use conn_max_idle_time instead"
```

sneako:
Looks great! I completely agree that we should update the telemetry event and variables as well to be consistent, good call

oliveigah:
Done! Let me know if I can improve the implementation some way! :smile:

lib/finch.ex
```diff
@@ -57,6 +57,25 @@ defmodule Finch do
...
       """,
       default: []
+    ],
+    pool_max_idle_time: [
+      type: :timeout,
+      doc: """
+      The maximum number of milliseconds that a pool can be idle before being terminated, used only by HTTP1 pools. \
+      This options is forwarded to NimblePool and it starts and idle verification cycle that may impact \
+      performance if misused. For instance setting a very low timeout may lead to pool restarts. \
+      Defaults to no timeout. \
```
sneako, line 67: NimbleOptions will automatically add a line that states the default, you can see here: https://hexdocs.pm/finch/Finch.html#start_link/1-pool-configuration-options
```suggestion
```

PR #176, opened 2022-04-07
Refactor telemetry events

### LINKED ISSUE #175: new unified telemetry event
I'm working on an OpenTelemetry library but I'm encountering some difficulties in the implementation while trying to capture various kinds of errors (e.g., `nxdomain` in the connect step). The underlying issue here is that there is no guaranteed 'start' and 'end' event when a request is executed in Finch. It requires a lot of error prone checks and edge-case handling to get it right. 

Would it make sense to emit a new set of telemetry events `[:finch, :execute_request, :start | :stop | :exception]` or just simply the `[:finch, :request_executed]` event? (with duration as measurement) This new event would basically wrap `request/3`.

dvic:
Fixes #175

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,10 @@
 ### Breaking changes
 - Telemetry updates #176
     - Rename the telemetry event `:request` to `:send` and `:response` to `:recv`.
-    - Replace `:scheme`, `:host`, `:port`, `:path`, `:method` fields in favor of a single `request` field.
-    - Rename the meta data field `:error` to `:reason` for all `:exception` events.
+    - Introduce a new `:request` field which contains the full `Finch.Request.t()` in place of the `:scheme`, `:host`, `:port`, `:path`, `:method` fields wherever possible. The new `:request` field can be found on the `:request`, `:queue`, `:send`, and `:recv` events.
+    - Rename the meta data field `:error` to `:reason` for all `:exception` events to follow the standard introduced in [telemetry](https://github.com/beam-telemetry/telemetry/blob/3f069cfd2193396bee221d0709287c1bdaa4fabf/src/telemetry.erl#L335)
     - Introduce a new `[:finch, :request, :start | :stop | :exception]` telemetry event that emits
-      whenever `Finch.request/3` or `Finch.stream/5` is called.
+      whenever `Finch.request/3` or `Finch.stream/5` are called.
 
 ## v0.11.0 (2022-03-28)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## v0.12.0 (unreleased)
+### Breaking changes
+- Telemetry updates #176
+    - Rename the telemetry event `:request` to `:send` and `:response` to `:recv`.
+    - Replace `:scheme`, `:host`, `:port`, `:path`, `:method` fields in favor of a single `request` field.
+    - Rename the meta data field `:error` to `:reason` for all `:exception` events.
+    - Introduce a new `[:finch, :request, :start | :stop | :exception]` telemetry event that emits
+      whenever `Finch.request/3` or `Finch.stream/5` is called.
+
 ## v0.11.0 (2022-03-28)
 
 - Add `:pool_max_idle_time` option to enable termination of idle HTTP/1 pools.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,15 @@
 # Changelog
 
 ## v0.12.0 (unreleased)
+
 ### Breaking changes
+
 - Telemetry updates #176
-    - Rename the telemetry event `:request` to `:send` and `:response` to `:recv`.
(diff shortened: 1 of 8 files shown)


dvic:
@sneako the only thing I'm wondering here is whether to use `error` or `reason` (or both) for the exception event. The `:telemetry` library uses `reason` by default but maybe for backwards compatible reasons we want to emit here also `error`? Or change the other events in a backwards compatible way to use `reason`?

sneako:
Ah good point. We should probably move towards the standard `reason`. Maybe we should update the existing exceptions to also include `reason` and note that the `error` fields are deprecated and will be removed in Finch 1.0.0?

dvic:
> Ah good point. We should probably move towards the standard `reason`. Maybe we should update the existing exceptions to also include `reason` and note that the `error` fields are deprecated and will be removed in Finch 1.0.0?

Sounds reasonable to me. Do you want to this in a separate PR?

And another thing that pop-ups in my mind: it would be nice if the OpenTelemetry library for Finch I'm building can support custom span names, e.g., `GET /users/:id` instead of the standard `HTTP GET` name (see https://github.com/ricardoccpaiva/opentelemetry_tesla/pull/17#issuecomment-1090781681). However, this would require the user to pass in the custom span name per request so that this info can be extracted when I'm consuming the `telemetry` events and creating the OpenTelemetry spans.

What do you think about adding a field called `private` in the `Finch.Request` struct to support this and other usecases?

lib/finch.ex
```diff
@@ -251,9 +251,17 @@ defmodule Finch do
...
-    {pool, pool_mod} = PoolManager.get_pool(name, shp)
-    pool_mod.request(pool, req, acc, fun, opts)
+    start_meta = %{request: req, name: name}
+
+    Finch.Telemetry.span(:stream, start_meta, fn ->
+      shp = build_shp(req)
+      {pool, pool_mod} = PoolManager.get_pool(name, shp)
+      result = pool_mod.request(pool, req, acc, fun, opts)
+
+      end_meta = Map.put(start_meta, :result, result)
```
keathley, line 261: Does it make sense to call this `:response`?

dvic: @sneako I came up with a different solution in 17ae19a (emit same event but make the result differ).

dvic:
I just pushed the next iteration :) It contains some doc fixes as well. Note that I also went ahead and refactored the metadata fields of the existing events to use a single `request` field where possible, I assume this is something that we want to do now in this breaking change as well?

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## v0.12.0 (unreleased)
+### Breaking changes
+- Telemetry updates #176
+    - Rename the telemetry event `:request` to `:send` and `:response` to `:recv`.
+    - Replace `:scheme`, `:host`, `:port`, `:path`, `:method` fields in favor of a single `request` field.
```
sneako, line 7:
```suggestion
    - Introduce a new `:request` field which contains the full `Finch.Request.t()` in place of the `:scheme`, `:host`, `:port`, `:path`, `:method` fields wherever possible. The new `:request` field can be found on the `:request`, `:queue`, `:send`, and `:recv` events.
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
+    - Rename the meta data field `:error` to `:reason` for all `:exception` events.
+    - Introduce a new `[:finch, :request, :start | :stop | :exception]` telemetry event that emits
+      whenever `Finch.request/3` or `Finch.stream/5` is called.
```
sneako, line 10:
```suggestion
      whenever `Finch.request/3` or `Finch.stream/5` are called.
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## v0.12.0 (unreleased)
+### Breaking changes
+- Telemetry updates #176
+    - Rename the telemetry event `:request` to `:send` and `:response` to `:recv`.
+    - Replace `:scheme`, `:host`, `:port`, `:path`, `:method` fields in favor of a single `request` field.
+    - Rename the meta data field `:error` to `:reason` for all `:exception` events.
```
sneako, line 8:
```suggestion
    - Rename the meta data field `:error` to `:reason` for all `:exception` events to follow the standard introduced in [telemetry](https://github.com/beam-telemetry/telemetry/blob/3f069cfd2193396bee221d0709287c1bdaa4fabf/src/telemetry.erl#L335)
```

sneako (changes requested):
Thanks @dvic ! Just a few minor suggestions, but overall I am happy with these changes.

Maybe we should also add a note to the README? Perhaps something like:
```
## Note: breaking changes have been introduced in 0.12.0 affecting the Telemetry events and their metadata. Please see the entry for 0.12.0 in CHANGELOG.md for further details.
```

dvic:
> 

Done :)

sneako (approved):
Thanks @dvic!

PR #181, opened 2022-04-14
Downgrade mime to 1.x in the lockfile

dvic:


dvic:
This PR attempts to fix the build for Elixir 1.7. Do we want to keep support for this version of Elixir? If not, then this PR can be closed :)

sneako:
Wow! I have been meaning to fix this and did not imagine it could be so simple. Thank you!!

PR #182, opened 2022-04-14
Add max response body size

eric0fw:
This is to avoid eating all the memory when fetching from untrusted links.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -294,6 +294,9 @@ defmodule Finch do
     * `:receive_timeout` - The maximum time to wait for a response before returning an error.
       Default value is `15_000`.
 
+    * `:max_response_size` - The maximum size of the response body. Exception raise has reason: `:response_body_too_large`.
+      Default value is `:infinity`.
+
   """
   @spec request(Request.t(), name(), keyword()) ::
           {:ok, Response.t()}
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -4,6 +4,7 @@ defmodule Finch.Conn do
   alias Mint.HTTP1
   alias Finch.Telemetry
   alias Finch.SSL
+  alias Finch.Error
 
   def new(scheme, host, port, opts, parent) do
     %{
@@ -93,9 +94,9 @@ defmodule Finch.Conn do
     end
   end
 
-  def request(%{mint: nil} = conn, _, _, _, _, _), do: {:error, conn, "Could not connect"}
+  def request(%{mint: nil} = conn, _, _, _, _, _, _), do: {:error, conn, "Could not connect"}
 
-  def request(conn, req, acc, fun, receive_timeout, idle_time) do
+  def request(conn, req, acc, fun, receive_timeout, idle_time, max_body) do
     full_path = Finch.Request.request_path(req)
 
     metadata = %{request: req}
@@ -111,7 +112,7 @@ defmodule Finch.Conn do
             {:ok, mint} ->
               Telemetry.stop(:send, start_time, metadata, extra_measurements)
               start_time = Telemetry.start(:recv, metadata, extra_measurements)
-              response = receive_response([], acc, fun, mint, ref, receive_timeout)
+              response = receive_response([], acc, fun, mint, ref, receive_timeout, max_body)
               handle_response(response, conn, metadata, start_time, extra_measurements)
 
             {:error, mint, error} ->
@@ -174,6 +175,11 @@ defmodule Finch.Conn do
         Telemetry.stop(:recv, start_time, metadata, extra_measurements)
         {:ok, %{conn | mint: mint}, acc}
 
+      {:error, _mint, error = :response_body_too_large} ->
+        metadata = Map.put(metadata, :error, error)
+        Telemetry.stop(:recv, start_time, metadata, extra_measurements)
+        {:error, close(conn), Error.exception(error)}
+
       {:error, mint, error} ->
         metadata = Map.put(metadata, :error, error)
         Telemetry.stop(:recv, start_time, metadata, extra_measurements)
@@ -181,26 +187,71 @@ defmodule Finch.Conn do
     end
   end
 
-  defp receive_response([], acc, fun, mint, ref, timeout) do
+  defp receive_response(entries, acc, fun, mint, ref, timeout, max_body, body_sum \\ 0)
+
+  defp receive_response([], acc, fun, mint, ref, timeout, max_body, body_sum) do
     case HTTP1.recv(mint, 0, timeout) do
       {:ok, mint, entries} ->
-        receive_response(entries, acc, fun, mint, ref, timeout)
+        receive_response(entries, acc, fun, mint, ref, timeout, max_body, body_sum)
 
       {:error, mint, error, _responses} ->
         {:error, mint, error}
     end
   end
 
-  defp receive_response([entry | entries], acc, fun, mint, ref, timeout) do
+  defp receive_response([entry | entries], acc, fun, mint, ref, timeout, max_body, body_sum) do
     case entry do
       {:status, ^ref, value} ->
-        receive_response(entries, fun.({:status, value}, acc), fun, mint, ref, timeout)
+        receive_response(
+          entries,
+          fun.({:status, value}, acc),
+          fun,
+          mint,
+          ref,
+          timeout,
+          max_body,
+          body_sum
+        )
 
       {:headers, ^ref, value} ->
-        receive_response(entries, fun.({:headers, value}, acc), fun, mint, ref, timeout)
+        with true <- is_integer(max_body),
+             {_header_name, content_length_str} <- List.keyfind(value, "content-length", 0),
+             {content_length, ""} = Integer.parse(content_length_str),
+             true <- content_length > max_body do
+          {:error, mint, :response_body_too_large}
+        else
+          _ ->
+            receive_response(
+              entries,
+              fun.({:headers, value}, acc),
+              fun,
+              mint,
+              ref,
+              timeout,
+              max_body,
+              body_sum
+            )
+        end
 
       {:data, ^ref, value} ->
-        receive_response(entries, fun.({:data, value}, acc), fun, mint, ref, timeout)
+        body_sum = :erlang.iolist_size(value) + body_sum
+
+        with true <- is_integer(max_body),
+             true <- body_sum > max_body do
+          {:error, mint, :response_body_too_large}
+        else
+          _ ->
+            receive_response(
+              entries,
+              fun.({:data, value}, acc),
+              fun,
+              mint,
+              ref,
+              timeout,
+              max_body,
+              body_sum
+            )
+        end
 
       {:done, ^ref} ->
         {:ok, mint, acc}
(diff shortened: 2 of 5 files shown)


lib/finch/http2/pool.ex
```diff
@@ -67,20 +76,76 @@ defmodule Finch.HTTP2.Pool do
 
           :gen_statem.call(pool, {:cancel, ref})
           clean_responses(ref)
-          Process.demonitor(monitor)
+          Process.demonitor(monitor, [:flush])
```
sneako, line 79: This seems correct but I am interested in what lead to you add it here but not above?

lib/finch/http1/conn.ex
```diff
@@ -174,33 +175,83 @@ defmodule Finch.Conn do
...
         Telemetry.stop(:recv, start_time, metadata, extra_measurements)
         {:error, %{conn | mint: mint}, error}
     end
   end
 
-  defp receive_response([], acc, fun, mint, ref, timeout) do
+  defp receive_response(entries, acc, fun, mint, ref, timeout, max_body, body_sum \\ 0)
+
+  defp receive_response([], acc, fun, mint, ref, timeout, max_body, body_sum) do
     case HTTP1.recv(mint, 0, timeout) do
```
sneako, line 193: The `0` passed in here is the number of bytes you want to accept from the socket. I believe a more correct implementation would pass a value here. See https://hexdocs.pm/mint/Mint.HTTP.html#recv/3

sneako (changes requested):
Hi @eric0fw thanks for taking the time to look in to this, and sorry for the delay, I took a short vacation :)

I agree that Finch users should be able to protect themselves in these situations.

I believe something similar could be achieved today by writing a custom streaming function, but I am interested in exploring ways to provide this functionality. 

It is very important that we make sure not to slow things down at all for users who are not interested in this functionality.

lib/finch/http1/conn.ex
```diff
@@ -174,33 +175,83 @@ defmodule Finch.Conn do
...
+              ref,
+              timeout,
+              max_body,
+              body_sum
+            )
+        end
 
       {:data, ^ref, value} ->
-        receive_response(entries, fun.({:data, value}, acc), fun, mint, ref, timeout)
+        body_sum = :erlang.iolist_size(value) + body_sum
```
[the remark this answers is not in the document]
eric0fw: I think this is debatable because the cost is tiny compared to the HTTP request itself, and the branch operation is likely to be more costly than the calculation ([pipeline stall](https://en.wikipedia.org/wiki/Pipeline_stall)). Also, I found it helpful when using `:recon_trace.calls` to track how much data was transferred by looking at the body_sum parameter.

lib/finch/http1/conn.ex
eric0fw: You could set an upper bound. Some hints are that the maximum binary size returned could match the kernel receive buffer size. Maybe `65_536` would be good enough since the overhead of a TCP or SSL socket would be similar in size.

lib/finch/http1/conn.ex
```diff
@@ -174,33 +175,83 @@ defmodule Finch.Conn do
...
 
+      {:error, _mint, error = :response_body_too_large} ->
+        metadata = Map.put(metadata, :error, error)
+        Telemetry.stop(:recv, start_time, metadata, extra_measurements)
+        {:error, close(conn), Error.exception(error)}
+
       {:error, mint, error} ->
         metadata = Map.put(metadata, :error, error)
         Telemetry.stop(:recv, start_time, metadata, extra_measurements)
         {:error, %{conn | mint: mint}, error}
```
eric0fw, line 186: Should you close the connection here? I closed it above because the subsequent request would fail if I didn't.

lib/finch/http2/pool.ex
```diff
@@ -48,14 +49,22 @@ defmodule Finch.HTTP2.Pool do
...
             metadata = Map.merge(metadata, %{status: status, headers: headers})
             Telemetry.stop(:recv, start_time, metadata)
             result
 
+          {:error, error = :response_body_too_large} ->
+            metadata = Map.put(metadata, :error, error)
+            Telemetry.stop(:recv, start_time, metadata)
+            :gen_statem.call(pool, {:cancel, ref})
+            clean_responses(ref)
+            Process.demonitor(monitor)
```
eric0fw, line 65: Add `:flush` parameter.

lib/finch.ex
```diff
@@ -294,6 +294,9 @@ defmodule Finch do
     * `:receive_timeout` - The maximum time to wait for a response before returning an error.
       Default value is `15_000`.
 
+    * `:max_response_size` - The maximum size of the response body. Exception raise has reason: `:response_body_too_large`.
```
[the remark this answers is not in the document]
eric0fw: NOTE: raise -> raised

sneako:
Hi @eric0fw ! I'm back from a short vacation and looking over this again.

I am hesitant to move forward here for a few reasons, but the main reason being that if we aren't going to pass the number of bytes that we want to receive from the socket to `Mint.HTTP.recv/3`, then this solution is not any better than what could be achieved with a custom streaming function.

This PR also includes changes that are orthogonal to the issue of adding a max response body size, which I would prefer to be added and thoroughly tested in a separate PR. Adding `[:flush]` to the `Process.demonitor/2` calls is one example, adding a call to `close/1` in `Conn.request/6` is another.

If you are under pressure at work to get this functionality today, I would suggest using a custom streaming function like this which should be very similar to the changes in this PR:

```elixir
      acc = {nil, [], [], 0}

      fun = fn
        {:status, value}, {_, headers, body, body_size_acc} ->
          {value, headers, body, body_size_acc}

        {:headers, value}, {status, headers, body, body_size_acc} -> 
          if content_size_too_large?(headers) do
            throw "body too large"
          else
            {status, headers ++ value, body, body_size_acc}
          end

        {:data, value}, {status, headers, body, body_size_acc} ->
          total_body_size = body_size_acc + byte_size(size)

          if total_body_size > max_body_size do
            throw "body too large"
          else
            {status, headers, [value | body], total_body_size}
          end
      end

     Finch.stream(req, name, acc, fun)
```

eric0fw:
> Hi @eric0fw ! I'm back from a short vacation and looking over this again.
> 
> I am hesitant to move forward here for a few reasons, but the main reason being that if we aren't going to pass the number of bytes that we want to receive from the socket to `Mint.HTTP.recv/3`, then this solution is not any better than what could be achieved with a custom streaming function.
> 
> This PR also includes changes that are orthogonal to the issue of adding a max response body size, which I would prefer to be added and thoroughly tested in a separate PR. Adding `[:flush]` to the `Process.demonitor/2` calls is one example, adding a call to `close/1` in `Conn.request/6` is another.
> 
> If you are under pressure at work to get this functionality today, I would suggest using a custom streaming function like this which should be very similar to the changes in this PR:
> 
> ```elixir
>       acc = {nil, [], [], 0}
> 
>       fun = fn
>         {:status, value}, {_, headers, body, body_size_acc} ->
>           {value, headers, body, body_size_acc}
> 
>         {:headers, value}, {status, headers, body, body_size_acc} -> 
>           if content_size_too_large?(headers) do
>             throw "body too large"
>           else
>             {status, headers ++ value, body, body_size_acc}
>           end
> 
>         {:data, value}, {status, headers, body, body_size_acc} ->
>           total_body_size = body_size_acc + byte_size(size)
> 
>           if total_body_size > max_body_size do
>             throw "body too large"
>           else
>             {status, headers, [value | body], total_body_size}
>           end
>       end
> 
>      Finch.stream(req, name, acc, fun)
> ```

If I include changes for the `recv` buffer and provide separate PRs for unrelated fixes. Will you take this PR?

lib/finch/http1/conn.ex
```diff
@@ -177,30 +178,71 @@ defmodule Finch.Conn do
       {:error, mint, error} ->
         metadata = Map.put(metadata, :error, error)
         Telemetry.stop(:recv, start_time, metadata, extra_measurements)
-        {:error, %{conn | mint: mint}, error}
+        {:error, close(%{conn | mint: mint}), error}
```
eric0fw, line 181: Move to other PR

lib/finch/http2/pool.ex
```diff
@@ -67,31 +76,87 @@ defmodule Finch.HTTP2.Pool do
...
+              monitor_ref,
+              fail_safe_timeout,
+              max_body,
+              body_sum
+            )
+        end
 
       {:done, ^ref} ->
-        Process.demonitor(monitor_ref)
+        Process.demonitor(monitor_ref, [:flush])
```
eric0fw, line 151: Move to separate PR

lib/finch/http2/pool.ex
```diff
@@ -67,31 +76,87 @@ defmodule Finch.HTTP2.Pool do
 
           :gen_statem.call(pool, {:cancel, ref})
           clean_responses(ref)
-          Process.demonitor(monitor)
+          Process.demonitor(monitor, [:flush])
```
eric0fw, line 79: Move to other PR

sneako:
Sure, I would still be willing to work together towards a solution that I would want to merge

eric0fw:
@sneako I made the changes requested, including the `recv` optimization.

lib/finch/http1/conn.ex
```diff
@@ -183,23 +189,25 @@ defmodule Finch.Conn do
...
     end
   end
 
-  defp receive_response(entries, acc, fun, mint, ref, timeout, status \\ nil, headers \\ [])
+  defp receive_response(entries, acc, fun, mint, ref, timeout, max_body, status \\ nil, headers \\ [], body_sum \\ 0)
 
-  defp receive_response([], acc, fun, mint, ref, timeout, status, headers) do
-    case MintHTTP1.recv(mint, 0, timeout) do
+  defp receive_response([], acc, fun, mint, ref, timeout, max_body, status, headers, body_sum) do
+    num_bytes_to_receive = if max_body === :infinity, do: 0, else: min(@default_tcp_recbuf_size, max_body - body_sum)
```
eric0fw, line 199: @sneako I'll run a simple experiment with `strace` in the next hour to determine what happens at the system call level if I set this to `max_body - body_sum` instead of calculating the minimum.

luiz-firework:
Hi, @sneako! I really think this is ready for merging. 

Can you provide any updates?

Thanks!

sneako:
Hi @luiz-firework, unfortunately there are still several issues I had commented on that have been 'marked as resolved' without any changes or comment...

Why do we need to add these additional calls to `close/1` and `Process.demonitor/1`? I would feel much better about adding these if there were some test cases that show they are necessary. Our current implementation is very heavily used and seems to work quite well, so I am not sure that they are necessary. Of course I would be happy to see some proof to the contrary, but 

Why are we only testing HTTP/2? We should definitely test both protocols.

I am also not such a fan of this particular usage of `with`, with the happy path happening in the `else` clause.

I would like to see the code to parse the content-length headers extracted in to a well named function, as well as the code that calculates the remaining max body length. Maybe this logic should be centralized in a new module so we can use it for both http/1 & 2?

The option name `:max_response_body` is long, but still feels a bit vague. Maybe it should be `:max_response_body_size` or `_length`?

sneako:
Closing due to inactivity, and I have created https://github.com/sneako/finch/issues/224 to track the issue

Issue #157, opened 2021-10-26
`Mint.UnsafeProxy` connection not handled in `Finch.Conn.request/6`


iwinux:
Hi, I'd like to report an issue of using Finch with HTTP proxy.

## Problem

Requests to plain HTTP (i.e.: non-HTTPS) URLs via proxy are handled in `Mint.UnsafeProxy`, but currently `Finch.Conn.request/6` is not aware of such case, resulting in `FunctionClauseError`.

## Reproduce

```elixir
defmodule FinchProxyTest.App do
  use Application

  def start(_type, _args) do
    children = [
      {Finch, name: FinchProxyTest, pools: pool_options()}
    ]

    Supervisor.start_link(children, strategy: :one_for_one)
  end

  def pool_options do
    %{host: host, port: port} = System.get_env("http_proxy", "") |> URI.parse()
    %{default: [protocol: :http1, conn_opts: [proxy: {:http, host, port, []}]]}
  end
end

# suppose env `http_proxy` is set
FinchProxyTest.Client.request "http://localhost:8000"
```

## Possible Solutions

* Handle `UnsafeProxy` connections in `Finch.HTTP1.Pool.request/5`

* Dynamic Pool Configs: allow users to pass in a callback instead of static pool configs (`fn (scheme, host, port) -> pool_config end`, for example)

Hajto:
@bsedat Merged a pr for this issue, shouldn't this issue be closed?

sneako:
Yes, thanks!

PR #184, opened 2022-04-20
Add support for Mint.UnsafeProxy connections

bsedat:
This adds several private proxy functions to Finch.HTTP1.Conn to dispatch to either Mint.UnsafeProxy or Mint.HTTP1. The dispatching is performed explicitly so as to better catch any changes to the underlying API with compiler errors/warnings where possible.

I don't think this completely finishes the scope of what was being discussed in #157 but it clears up the runtime errors and now handles HTTP 1.1 proxy configuration.

### DIFF
--- a/lib/finch/http1/mint_http1.ex
+++ b/lib/finch/http1/mint_http1.ex
@@ -1,0 +1,74 @@
+defmodule Finch.MintHTTP1 do
+  @moduledoc false
+
+  alias Mint.HTTP1
+  alias Mint.UnsafeProxy
+
+  def connect(scheme, host, port, conn_opts) do
+    Mint.HTTP.connect(scheme, host, port, conn_opts)
+  end
+
+  def open?(%HTTP1{} = mint) do
+    HTTP1.open?(mint)
+  end
+
+  def open?(%UnsafeProxy{} = mint) do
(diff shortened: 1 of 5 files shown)


sneako:
Thanks @bsedat, this looks good to me!

If you don't mind could you just please reorganize this module a bit? I would like to keep the public functions above the private functions.

That plus a quick `mix format` and this should be good to go.

bsedat:
I moved the functions to a new module because they all had the same prefix, but let me know if you prefer keeping them in `Finch.Conn` and I can move them back and reorganize as requested.

lib/finch/http1/mint_http1.ex
```diff
@@ -0,0 +1,74 @@
+defmodule Finch.MintHTTP1 do
+  @moduledoc false
+
```
sneako, line 3:
```suggestion
  @moduledoc false
  
  # This module exists to ensure that protocol negotiation does not accidentally cause Mint to open
  # HTTP2 connections in our HTTP1 pools.

```

sneako:
Thanks @bsedat ! Let's just add this comment as a reminder as to why we have this new module, then I would say this is good to go

sneako:
Thank you!

PR #185, opened 2022-04-23
Fix and improve Telemetry doc formatting

joshprice:
- Fixed a missing ending backtick which broke formatting
- Used backticks to highlight code items for clarity

PR #187, opened 2022-04-25
Improve formatting of Telemetry docs (part 2)

joshprice:
- Fix the other missing backtick I didn't spot in #185
- Remove colons and add newlines for consistency thoughout

### DIFF
--- a/lib/finch/telemetry.ex
+++ b/lib/finch/telemetry.ex
@@ -2,7 +2,7 @@ defmodule Finch.Telemetry do
   @moduledoc """
   Telemetry integration.
 
-  Unless specified, all time's are in `:native` units.
+  Unless specified, all times are in `:native` units.
 
   Finch executes the following events:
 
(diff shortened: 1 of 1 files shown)


sneako:
Thank you! This time I checked out the PR and generated the docs :smile:  Can we get the "Measurements" and "Metadata" titles aligned at the same level of indentation?
telemetry-docs

joshprice:
Weird, I didn't change those, so assuming it was already like that. Will take a look.

joshprice:
OK not as many options here as I thought there might be. I don't think you can nest like that, although it's possibly an Earmark bug?

How do you feel about this as a workaround? I reckon taking the bullet off the event looks better, but your call.

Screen Shot 2022-04-25 at 7 45 31 pm

joshprice:
Pushed up the fix, let me know if you want to do a diff way. Might be worth just double checking all the new h3 titles, just in case I missed something.

sneako:
Looks good to me, thank you!

PR #188, opened 2022-04-25
Improve telemetry docs

joshprice:
- Improve some grammar
- Capitalise bullets consistently
- Capitalise PID
- Move deprecated `max_idle_time_exceeded` event to bottom and mark more clearly as deprecated

### DIFF
--- a/lib/finch/telemetry.ex
+++ b/lib/finch/telemetry.ex
@@ -25,7 +25,7 @@ defmodule Finch.Telemetry do
 
   #### Measurements
 
-  * `:duration` - Duration how long the request took.
+  * `:duration` - Time taken from the request start event.
 
   #### Metadata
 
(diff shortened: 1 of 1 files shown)


joshprice:
Not sure what's causing that test failure - flaky test?

sneako:
Ah, sorry for the flakiness! What a silly mistake I made, and I'm surprised it took this long to show up. We had been using `__ENV__.function` to get a unique existing atom that we can use as the Finch `:name` in each test. I extracted that in to a function just to clean things up a bit, but now that means, `__ENV__.function` always returns `:finch_name` :upside_down_face:

sneako:
Looks like we could use this macro instead of the current `finch_name/0` function:

```
  defmacrop finch_name do
    {name, _} = __CALLER__.function
    name
  end
```

joshprice:
Did you already make a PR? Happy to do one quickly

sneako:
Thank you!

PR #189, opened 2022-04-25
Ensure test name is unique in finch_name helper

joshprice:
This was causing a failing test in #188 due to test names not being unique.

### DIFF
--- a/test/finch_test.exs
+++ b/test/finch_test.exs
@@ -7,6 +7,15 @@ defmodule FinchTest do
   alias Finch.Response
   alias Finch.MockSocketServer
 
+  # returns the name of the current test as an atom which can be used as the Finch name
+  # to enable async tests with many different instances of Finch running
+  # this needs to be macro to retain the caller function's name
+  defmacrop finch_name do
+    {name, _} = __CALLER__.function
+    IO.inspect(name)
+    name
+  end
+
   setup do
     {:ok, bypass: Bypass.open()}
   end
@@ -1082,12 +1091,5 @@ defmodule FinchTest do
 
   defp expect_any(bypass) do
     Bypass.expect(bypass, fn conn -> Plug.Conn.send_resp(conn, 200, "OK") end)
-  end
-
-  # returns the name of the current test as an atom which can be used as the Finch name
-  # to enable async tests with many different instances of Finch running
-  defp finch_name do
-    {name, _} = __ENV__.function
-    name
   end
 end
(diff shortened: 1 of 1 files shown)


test/finch_test.exs
```diff
@@ -7,6 +7,15 @@ defmodule FinchTest do
   alias Finch.Response
   alias Finch.MockSocketServer
 
+  # returns the name of the current test as an atom which can be used as the Finch name
+  # to enable async tests with many different instances of Finch running
+  # this needs to be macro to retain the caller function's name
+  defmacrop finch_name do
+    {name, _} = __CALLER__.function
```
wojtekmach, line 14: Nice catch! It's a bigger change but perhaps we should just do it explicitly?

```diff
-    test "max_idle_time is deprecated" do
+    test "max_idle_time is deprecated", c do
       msg =
         capture_io(:stderr, fn ->
           start_supervised!(
-            {Finch, name: finch_name(), pools: %{default: [max_idle_time: 1_000]}}
+            {Finch, name: c.test, pools: %{default: [max_idle_time: 1_000]}}
           )
         end)
```

etc!

joshprice: Also not my catch, it was all @sneako ;)

sneako: Ah weird that test is still failing. I can try to take a closer look this evening

joshprice: I think I might have an idea. The failing test is using `finch_name()` in the setup and `__ENV__.function` in the test:

```
  describe "telemetry" do
    setup %{bypass: bypass} do
      Bypass.expect_once(bypass, "GET", "/", fn conn ->
        Plug.Conn.send_resp(conn, 200, "OK")
      end)

      client = finch_name()
      start_supervised!({Finch, name: client})

      {:ok, client: client}
    end

    test "reports request spans", %{bypass: bypass, client: client} do
      {test_name, _arity} = __ENV__.function |> IO.inspect()
```

joshprice: The test that is failing now is not the same as the test that failed previously. This was the flaky test, that presumably just happened to collide previously:

```
1) test request/3 successful post request, with body and query string (FinchTest)
Error:      test/finch_test.exs:201
     ** (RuntimeError) failed to start child with the spec {Finch, [name: :finch_name]}.
     Reason: already started: #PID<0.1074.0>
     code: start_supervised!({Finch, name: finch_name()})
     stacktrace:
       (ex_unit 1.12.3) lib/ex_unit/callbacks.ex:[43](https://github.com/sneako/finch/runs/6155762964?check_suite_focus=true#step:13:43)3: ExUnit.Callbacks.start_supervised!/2
```

sneako:
Closing in favor of https://github.com/sneako/finch/pull/190, thanks @joshprice!

PR #190, opened 2022-04-25
Test Refactor

sneako:
- introduce `FinchCase`
- use test name from context to provide a unique Finch :name
- extract all Telemetry tests to a new test file (we had multiple describe blocks for telemetry tests in the huge FinchTest file)

### DIFF
--- a/test/finch/http1/pool_test.exs
+++ b/test/finch/http1/pool_test.exs
@@ -1,15 +1,9 @@
 defmodule Finch.HTTP1.PoolTest do
-  use ExUnit.Case
+  use FinchCase, async: true
 
-  setup do
-    {:ok, bypass: Bypass.open()}
-  end
-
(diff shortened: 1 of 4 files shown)


sneako:
What do you think @joshprice and @wojtekmach ?

I used the test name from the context, but I stuck it in a `:finch_name` key because I thought that would be more clear than sprinkling `c.test` everywhere.

sneako:
Huh weird that tests are failing on Elixir 1.7, looking in to that…

sneako:
Also, we should probably add Telemetry tests for H2 pools as well...

wojtekmach:
> because I thought that would be more clear than sprinkling c.test everywhere.

👍

sneako:
The Telemetry test flakiness has been resolved by setting that file to `async: false`.

PR #194, opened 2022-05-02
Return `Finch.Error` instead of propagating `Mint.TransportError`

sato11:
While I was giving a try to solve https://github.com/sneako/finch/issues/186 I found this unit of changes might be independently testable and reviewable. So let me submit this first. What do you think?

---

Replaced `with` with `case` to roughly catch error.
This can at least stop propagating `Mint.TransportError`.

For other sorts of errors, it just returns themselves.
Since `Finch.Error` wants the reason given to be atom,
I've refrained from trying to pack everything into `Finch.Error`.

before:

```elixir
iex(2)> Finch.build(:get, "https://") |> Finch.request(MyFinch)
{:error,
 %Mint.TransportError{
   reason: {:options,
    {:socket_options,
     [
       nodelay: true,
       keepalive: true,
       packet_size: 0,
       packet: 0,
       header: 0,
       active: false,
       mode: :binary
     ]}}
 }}
```

after:

```elixir
iex(29)> Finch.build(:get, "https://") |> Finch.request(MyFinch)
{:error,
 %Finch.Error{
   reason: {:options,
    {:socket_options,
     [
       nodelay: true,
       keepalive: true,
       packet_size: 0,
       packet: 0,
       header: 0,
       active: false,
       mode: :binary
     ]}}
 }}
```

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -5,7 +5,7 @@ defmodule Finch do
              |> String.split("<!-- MDOC !-->")
              |> Enum.fetch!(1)
 
-  alias Finch.{PoolManager, Request, Response}
+  alias Finch.{PoolManager, Request, Response, Error}
 
   use Supervisor
 
@@ -310,13 +310,21 @@ defmodule Finch do
         {:data, value}, {status, headers, body} -> {status, headers, [value | body]}
       end
 
-      with {:ok, {status, headers, body}} <- __stream__(req, name, acc, fun, opts) do
-        {:ok,
-         %Response{
-           status: status,
-           headers: headers,
-           body: body |> Enum.reverse() |> IO.iodata_to_binary()
-         }}
+      case __stream__(req, name, acc, fun, opts) do
+        {:ok, {status, headers, body}} ->
+          {:ok,
+           %Response{
+             status: status,
+             headers: headers,
+             body: body |> Enum.reverse() |> IO.iodata_to_binary()
+           }}
+
+        {:error, %{reason: reason}} ->
+          {:error, %Error{reason: reason}}
+
+        # If something unexpected has happened, returns it.
+        x ->
+          x
       end
     end
   end
--- a/lib/finch/error.ex
+++ b/lib/finch/error.ex
@@ -2,7 +2,7 @@ defmodule Finch.Error do
   @moduledoc """
   An HTTP error.
 
-  This exception struct is used to represent errors of all sorts for the HTTP/2 protocol.
+  This exception struct is used to represent errors of all sorts.
   """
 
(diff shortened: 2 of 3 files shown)


wojtekmach:
> This can at least stop propagating Mint.TransportError.

I believe there's nothing wrong with Finch returning exceptions other than `Finch.Error`. The only promise Finch makes is that on errors it returns `{:error, Exception.t}`. That is, any exception struct, not any particular one.

I think it is desirable to keep `Mint.TransportError` and others as they might have their own preferred way of displaying errors (the Exception.message/1 callback). Here's an example:

```elixir
iex> Finch.start_link(name: MyFinch); r = Finch.build(:get, "http://bad") ; {:error, e} = Finch.request(r, MyFinch); raise e
** (Mint.TransportError) non-existing domain
```

on this branch we would have gotten: 

```
** (FinchError) :nxdomain
```

wojtekmach:
> This can at least stop propagating Mint.TransportError.

I believe there's nothing wrong with Finch returning exceptions other than `Finch.Error`. The only promise Finch makes is that on errors it returns `{:error, Exception.t}`. That is, any exception struct, not any particular one.

I think it is desirable to keep `Mint.TransportError` and others as they might have their own preferred way of displaying errors (the Exception.message/1 callback). Here's an example:

```elixir
iex> Finch.start_link(name: MyFinch); r = Finch.build(:get, "http://bad") ; {:error, e} = Finch.request(r, MyFinch); raise e
** (Mint.TransportError) non-existing domain
```

on this branch we would have gotten: 

```
** (FinchError) :nxdomain
```

wojtekmach:
This is not really related to this particular change per se but:

> ```
> iex(29)> Finch.build(:get, "https://") |> Finch.request(MyFinch)
> {:error,
>  %Finch.Error{
>    reason: {:options,
>     {:socket_options,
>      [
>        nodelay: true,
>        keepalive: true,
>        packet_size: 0,
>        packet: 0,
>        header: 0,
>        active: false,
>        mode: :binary
>      ]}}
>  }}
> ```

but this pretty unfortunate example because this is not at all when we want to have returned in this particular case. I think what is happening is due to invalid URL, Mint is somehow doing an incorrect `:ssl.connect` call. I'd argue that by wrapping the error in Finch.Error struct, it makes it a bit harder to debug it.

sato11:
Thanks for the comment, @wojtekmach!

Well, sorry that I've misunderstood the premises in the first place, as well as I've misused the term propagation. 😭 I believed it was nicer but after reading your feedback I'm convinced it's fine to let it be. Regarding `Exception.message/1`, I was not quite aware of how to take advantage of it but now I believe I am. 

Thanks again for the lesson, I much appreciate it 🙂

PR #196, opened 2022-05-03
accumulate headers and status code for telemetry events

### LINKED ISSUE #192: Finch.stream/5 crashes
Hello!

The following code crashes:

```
iex(1)> fun = fn _msg, :ok -> :ok end
#Function<43.65746770/2 in :erl_eval.expr/5>
iex(1)> Finch.build(:get, "https://elixir-lang.org/") |> Finch.stream(MyFinch, :ok, fun)
** (CaseClauseError) no case clause matching: {:ok, %Mint.HTTP1{buffer: "", host: "elixir-lang.org", mode: :passive, port: 443, private: %{}, proxy_headers: [], request: nil, requests: {[], []}, scheme_as_string: "https", socket: {:sslsocket, {:gen_tcp, #Port<0.24>, :tls_connection, :undefined}, [#PID<0.977.0>, #PID<0.976.0>]}, state: :open, streaming_request: nil, transport: Mint.Core.Transport.SSL}, :ok}
    (finch 0.11.0) lib/finch/http1/conn.ex:178: Finch.Conn.handle_response/5
    (finch 0.11.0) lib/finch/http1/conn.ex:122: Finch.Conn.request/6
    (finch 0.11.0) lib/finch/http1/pool.ex:50: anonymous fn/8 in Finch.HTTP1.Pool.request/5
    (nimble_pool 0.2.6) lib/nimble_pool.ex:349: NimblePool.checkout!/4
    (finch 0.11.0) lib/finch/http1/pool.ex:43: Finch.HTTP1.Pool.request/5
```

The code expects that the stream function returns `{status, header, body}`. I saw that the `request/3` reuses the stream function and it uses this tuple. I guess this is the origin of the problem. I can help if you want, but I don't know how do you want to solve the problem.

Thank you for your time

sneako:
It is unfortunate that we are building the list of headers twice in the case where the default streaming function is used, but I doubt the performance impact will be significant.

I am opening this as a draft while I test for performance regressions. Will report back here shortly.

cc @dvic 

Closes #192

### DIFF
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -175,44 +175,73 @@ defmodule Finch.Conn do
 
   defp handle_response(response, conn, metadata, start_time, extra_measurements) do
     case response do
-      {:ok, mint, {status, headers, _} = acc} ->
+      {:ok, mint, acc, {status, headers}} ->
         metadata = Map.merge(metadata, %{status: status, headers: headers})
         Telemetry.stop(:recv, start_time, metadata, extra_measurements)
         {:ok, %{conn | mint: mint}, acc}
 
-      {:error, mint, error} ->
-        metadata = Map.put(metadata, :error, error)
+      {:error, mint, error, {status, headers}} ->
+        metadata = Map.merge(metadata, %{error: error, status: status, headers: headers})
         Telemetry.stop(:recv, start_time, metadata, extra_measurements)
         {:error, %{conn | mint: mint}, error}
     end
   end
 
-  defp receive_response([], acc, fun, mint, ref, timeout) do
+  defp receive_response(entries, acc, fun, mint, ref, timeout, status \\ nil, headers \\ [])
+
+  defp receive_response([], acc, fun, mint, ref, timeout, status, headers) do
     case MintHTTP1.recv(mint, 0, timeout) do
       {:ok, mint, entries} ->
-        receive_response(entries, acc, fun, mint, ref, timeout)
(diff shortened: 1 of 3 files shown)


lib/finch/http1/conn.ex
```diff
@@ -169,44 +169,73 @@ defmodule Finch.Conn do
...
-        receive_response(entries, fun.({:headers, value}, acc), fun, mint, ref, timeout)
+        receive_response(
+          entries,
+          fun.({:headers, value}, acc),
+          fun,
+          mint,
+          ref,
+          timeout,
+          status,
+          headers ++ value
```
dvic, line 219: I'm not familiar with the mint internals but can this message received more than once? If not then this can be just `value` right? (instead of `headers ++ value`)

sneako: I believe that we could see multiple `{:headers, ...}` tuples if a response had trailing headers in addition to standard headers. I am looking at how I could write a test case for that now...

sneako: It appears that while Cowboy does support trailing headers, Plug.Cowboy does not yet allow you to send them.

sneako:
Performance seems consistent with the previous version, but I'll let this test run for a few more hours.

lib/finch/http1/conn.ex
dvic: Ahh too bad..

dvic:
> 

👍 Nice!

PR #202, opened 2022-07-24
Define Finch.child_spec/1

sneako:
By default it will set the child_spec :id to the Finch :name. This
allows multiple Finch instances to be started under a single supervisor
without any additional configuration.

See https://github.com/phoenixframework/phoenix/issues/4876#issuecomment-1193323397 for discussion.

cc @wojtekmach


wojtekmach (approved):
❤️

PR #205, opened 2022-11-02
Add nimble_options 0.5

philippeqd:
Relaxes the nimble_options package version by adding version 0.5 since I got version conflicts when using both finch 0.13 and nimble_options 0.5 in my project.


philippeqd:
I noticed this issue has already been solved in: https://github.com/sneako/finch/commit/711e0d49278c1b11da90191116d762d366b7e74d

PR #210, opened 2023-02-17
Allow using with castore 1.0

michallepicki:
CAStore 1.0 was released with no major API changes.

Alternatively the requirement could be just bumped to `1.0`.

I didn't update the `mix.lock` file, because `mint` also depends on `castore`, and because of some [changes](https://github.com/elixir-mint/mint/pull/331) tests would fail here after updating `mint`.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -34,7 +34,7 @@ defmodule Finch.MixProject do
   defp deps do
     [
       {:mint, "~> 1.3"},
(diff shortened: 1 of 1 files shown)


sneako:
Thank you!

AntoineAugusti:
Thanks for this PR ❤️ We @ transport.data.gouv.fr are looking forward to a release including this change 🙏

PR #221, opened 2023-04-13
Fully qualify call to request function

jwien001:
In `request!`, call `request` using the module name.

This fixes cases where application code calls `request!` while test code mocks `request`. In those cases, `request!` would call the original, unmocked `request`, leading to unexpected behavior. For example, using an ExVCR cassette on code that calls `request!` won't get the cassette response. By using the fully qualified module name, `request!` will call the mocked `request` instead.


sneako:
Sorry, but this sounds like exVCR needs to be updated to support the new request! function. I don't think Finch should make non idiomatic changes just to accommodate this particular testing library.

Issue #26, opened 2020-04-20
Fix timeouts


sneako:
Right now, we pass the same value for receive_timeout, everywhere that will accept it, with no regard for any time that has already elapsed. We need update the receive_timeouts we pass by subtracting the elapsed time of any intermediate steps in order to properly respect the timeout.

josevalim:
We may want to have two timeouts: one for each individual read/receive from the socket and one for the whole connect+request+response. Or even three: connect, individual read/receive, and whole request+response.

sneako:
We also have the pool timeout as well, so there could be a total of 4 timeouts :) It seems like a lot but I suppose most users would just keep the defaults and only configure the ones they actually care about.

whatyouhide:
I think having sane defaults makes up for it. I would love to be able to have this much control. :)

sneako:
I just merged #233 which addresses the original issue, where it was possible that the elapsed time experienced by the user could greatly exceed the configured receive_timeout. 

I will leave this issue open, because I would still be willing to accept changes to make the timeouts more configurable as discussed above.

PR #223, opened 2023-04-23
Recursively update receive_timeout with elapsed time

louiscb:
Solves part of issue #26 by recursively reducing the `receive_timeout` value with the elapsed time from Mint. This is only an issue present in http1 and not http2 due to the `receive -> after` code here https://github.com/sneako/finch/blob/main/lib/finch/http2/pool.ex#L132

The new test in this PR should fail in the current Finch source code as the chunked response will cause the `receive_response/8`  function to be recursively called with the same `receive_timeout`. In this case the actual time experienced by a user making a request will be several times longer than their specified `receive_timeout`.

### DIFF
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -190,8 +190,12 @@ defmodule Finch.Conn do
   defp receive_response(entries, acc, fun, mint, ref, timeout, status \\ nil, headers \\ [])
 
   defp receive_response([], acc, fun, mint, ref, timeout, status, headers) do
+    start_time = System.monotonic_time(:millisecond)
+
     case MintHTTP1.recv(mint, 0, timeout) do
       {:ok, mint, entries} ->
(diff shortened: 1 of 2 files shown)


ericmj (approved):
👍

louiscb:
Tested the PR in a high load environment. Turns out it is possible to produce an error from a race condition where a negative timeout is passed into Mint. So I added a guard clause that checks if the timeout is < 0. 

Stacktrace:

```
no function clause matching in :ssl.recv/3
ssl.erl in :ssl.recv/3 at line 887
lib/mint/core/transport/ssl.ex in Mint.Core.Transport.SSL.recv/3 at line 378
lib/mint/http1.ex in Mint.HTTP1.recv/3 at line 500
lib/finch/http1/conn.ex in Finch.Conn.receive_response/8 at line 195
```

sneako:
Thanks!

PR #227, opened 2023-05-25
Discard not ready connections from HTTP/2 pool

### LINKED ISSUE #216: Discussion: should HTTP/2 pool skip over not ready connections?
We sometimes encounter this error in our production release:
```elixir
{:error, %Finch.Error{reason: :read_only}}
```

This error indicates that the checked out pool is close for writing and only able to read responses. If the pool is in `:disconnected` or `:connected_read_only` state, it's not able to make requests (I'll refer these two states as `not_ready` from now on).

My point is we should be able to filter out `not_ready` pools while [performing lookup](https://github.com/sneako/finch/blob/main/lib/finch/pool_manager.ex#L43). Right now, there's no way to efficiently get pool's current state other than sending a message to it (GenServer.call, :sys.get_state, ...). How about lifting the pool state to a ets table?

Goose97:
HTTP2 connections will unregister themselves from the pool once enter connected_read_only and disconnected states, and register themselves once they reconnect.

Fixes #216

### DIFF
--- a/lib/finch/http2/pool.ex
+++ b/lib/finch/http2/pool.ex
@@ -161,8 +161,6 @@ defmodule Finch.HTTP2.Pool do
 
   @impl true
   def init({{scheme, host, port} = shp, registry, _pool_size, pool_opts}) do
-    {:ok, _} = Registry.register(registry, shp, __MODULE__)
-
     data = %{
       conn: nil,
       scheme: scheme,
@@ -171,7 +169,8 @@ defmodule Finch.HTTP2.Pool do
       requests: %{},
       backoff_base: 500,
       backoff_max: 10_000,
-      connect_opts: pool_opts[:conn_opts] || []
+      connect_opts: pool_opts[:conn_opts] || [],
+      registry: registry
     }
 
     {:ok, :disconnected, data, {:next_event, :internal, {:connect, 0}}}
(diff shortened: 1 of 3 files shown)


lib/finch/http2/pool.ex
```diff
@@ -161,8 +161,6 @@ defmodule Finch.HTTP2.Pool do
 
   @impl true
   def init({{scheme, host, port} = shp, registry, _pool_size, pool_opts}) do
-    {:ok, _} = Registry.register(registry, shp, __MODULE__)
```
Goose97, line 164: I remove the registering code here since we have another one after we enter connected state. IMO, registration happens after we establish the connection makes more sense

sneako:
Thanks for taking this on! I am curious, did you run any kind of stress testing on this? I would like to see how the error rate and throughput compare to what we have in main.

sneako:
Sorry for the delay, but I finally got around to testing this out and found just a minimal performance regression but a total elimination of errors that are present on finch:main, so I think we should proceed with this change. I fixed up a couple things, but I want to merge #228 first,  and then deal with any merge conflicts in this pr.

hkrutzer:
@sneako is this the last issue for a new release?

nallwhy:
I made a application using OpenAI API via Finch HTTP2 pool.
And I got so much `(Finch.Error) disconnected` errors.
Is this issue related to the problem?

sneako:
> I made a application using OpenAI API via Finch HTTP2 pool.
> 
> And I got so much `(Finch.Error) disconnected` errors.
> 
> Is this issue related to the problem?

It could be. I would be very interested to know if this PR helps with the issue you are seeing. I have been struggling to find time to test this properly myself.

Issue #208, opened 2022-12-28
Lazy streaming 


dmkl:
Is it possible to handle a get request as a lazy stream, processing it chunk by chunk "externally"? Something like

```elixir
finch_stream
|> Stream.map(&handle_each_chunk/1)
|> Stream.run()
```

So, it will pull chunks one by one from outside of the `finch_stream`. I'm not sure if it's possible to use `%Stream{}` in the accumulator function of `Finch.stream' or wrap it around `Stream.resource` somehow.

hkrutzer:
I would also like this feature so I can perform a request, do some processing and send the result out immediately with `Plug.conn.send_chunked`, without having to buffer anything as that would take a lot of memory.

It doesn't seem possible with the current API, but Mint should be able to do it in passive mode with `Stream.resource` or with a buffer process and `Enumerable.t()` I think. @josevalim as you wrote the original implementation in https://github.com/sneako/finch/pull/59, do you call if there was a specific reason for using an accumulator instead of producing a stream, or did it just fit the use case more at the time?

sneako:
In the current state, it is not possible to get a Stream out of Finch. Finch was designed for hammering a static set of hosts with a high rate of requests (we had this use case where I work, as well as where Keathley was working at the time). For this reason, the library is focused on connection reuse and limiting copying whenever possible. We have extended the feature set a bit where it has been possible to do so without adding overhead, but it was never intended to be a general purpose, full-featured client. 

@hkrutzer I'm probably missing something, and I have not worked with `Plug.Conn.send_chunked/2` myself, but why wouldn't something like this work?

```elixir
      fun = fn
        {:status, _value}, conn ->
          conn

        {:headers, _value}, conn ->
          conn

        {:data, value}, conn ->
          processed = process(value)

          case Plug.Conn.chunk(conn, processed) do
            {:ok, conn} -> conn
            {:error, _} -> raise "failed to send chunk"
          end
      end

      conn = Plug.Conn.send_chunked(conn, :ok)

      Finch.build(:get, "http://example.com")
      |> Finch.stream(finch_name, conn, fun)
```

hkrutzer:
[...]

PR #228, opened 2023-06-04
Add support for async requests

zachallaun:
See #208 for context and past discussion.

---

This PR is incomplete, but I think enough is here to share and get some feedback.

### Todo

- [ ] `Finch.HTTP2.Pool` - correctly send errors that are currently sent as `:gen_statem` replies in `continue_requests`
- [ ] `Finch.HTTP2.Pool` - add tests for `:connected_read_only` state
- [ ] `Finch.HTTP2.Pool` - fix significant duplication between `:request` call and `:async_request` cast
- [ ] Add/improve documentation
- [ ] Benchmark against `main` to ensure no performance regressions for sync requests

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -359,4 +359,34 @@ defmodule Finch do
       {:error, exception} -> raise exception
     end
   end
+
+  @opaque request_ref() :: Task.t()
+
+  @doc """
+  Sends an HTTP request asynchronously, returning a request reference.
+
+  Response information will be sent as messages to the process that
+  issued the request.
+  """
+  @spec async_request(Request.t(), name(), keyword()) :: request_ref()
+  def async_request(%Request{} = req, name, opts \\ []) do
+    task = Task.async(__MODULE__, :__async_request__, [req, name, opts, self()])
+    send(task.pid, {:task, task})
+    task
+  end
+
+  @doc false
+  def __async_request__(req, name, opts, owner) do
+    task = receive do: ({:task, task} -> task)
+
+    case stream(req, name, {owner, task}, &send_chunk/2, opts) do
+      {:ok, _} -> send(owner, {task, :done})
+      {:error, exception} -> send(owner, {task, {:error, exception}})
+    end
+  end
+
+  defp send_chunk(chunk, {owner, task}) do
+    send(owner, {task, chunk})
+    {owner, task}
+  end
 end
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -115,6 +115,7 @@ defmodule Finch do
       registry_name: name,
       manager_name: manager_name(name),
       supervisor_name: pool_supervisor_name(name),
+      task_supervisor_name: task_supervisor_name(name),
       default_pool_config: default_pool_config,
       pools: pools
     }
@@ -133,6 +134,7 @@ defmodule Finch do
   def init(config) do
     children = [
       {DynamicSupervisor, name: config.supervisor_name, strategy: :one_for_one},
+      {Task.Supervisor, name: config.task_supervisor_name},
       {Registry, [keys: :duplicate, name: config.registry_name, meta: [config: config]]},
       {PoolManager, config}
     ]
@@ -223,6 +225,7 @@ defmodule Finch do
   defp supervisor_name(name), do: :"#{name}.Supervisor"
   defp manager_name(name), do: :"#{name}.PoolManager"
   defp pool_supervisor_name(name), do: :"#{name}.PoolSupervisor"
+  defp task_supervisor_name(name), do: :"#{name}.TaskSupervisor"
 
   defmacrop request_span(request, name, do: block) do
     quote do
@@ -360,7 +363,7 @@ defmodule Finch do
     end
   end
 
-  @opaque request_ref() :: Task.t()
+  @opaque request_ref() :: {pid(), reference()}
 
   @doc """
   Sends an HTTP request asynchronously, returning a request reference.
@@ -370,23 +373,43 @@ defmodule Finch do
   """
   @spec async_request(Request.t(), name(), keyword()) :: request_ref()
   def async_request(%Request{} = req, name, opts \\ []) do
-    task = Task.async(__MODULE__, :__async_request__, [req, name, opts, self()])
-    send(task.pid, {:task, task})
-    task
+    # TODO: get supervisor from config when this moves into the adapter
+    sup = task_supervisor_name(name)
+
+    %{pid: pid, ref: ref} =
+      Task.Supervisor.async_nolink(sup, __MODULE__, :__async_request__, [self(), req, name, opts])
+
+    send(pid, {:ref, ref})
+
+    {pid, ref}
   end
 
   @doc false
-  def __async_request__(req, name, opts, owner) do
-    task = receive do: ({:task, task} -> task)
+  def __async_request__(owner, req, name, opts) do
+    ref = receive do: ({:ref, ref} -> ref)
+    request_ref = {self(), ref}
 
-    case stream(req, name, {owner, task}, &send_chunk/2, opts) do
-      {:ok, _} -> send(owner, {task, :done})
-      {:error, exception} -> send(owner, {task, {:error, exception}})
+    case stream(req, name, {owner, request_ref}, &send_chunk/2, opts) do
+      {:ok, _} -> send(owner, {request_ref, :done})
+      {:error, exception} -> send(owner, {request_ref, {:error, exception}})
     end
   end
 
-  defp send_chunk(chunk, {owner, task}) do
-    send(owner, {task, chunk})
-    {owner, task}
+  defp send_chunk(chunk, {owner, request_ref}) do
+    send(owner, {request_ref, chunk})
+    {owner, request_ref}
+  end
+
+  @doc """
+  Cancels a request sent with `async_request/3`.
+  """
+  @spec cancel_async_request(request_ref(), name()) :: :ok | {:error, :not_found}
+  def cancel_async_request(request_ref, name) do
+    # TODO: get supervisor from config when this moves into the adapter
+    sup = task_supervisor_name(name)
+
+    {pid, ref} = request_ref
+    Process.demonitor(ref, [:flush])
+    Task.Supervisor.terminate_child(sup, pid)
   end
 end
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -6,6 +6,7 @@ defmodule Finch do
              |> Enum.fetch!(1)
 
   alias Finch.{PoolManager, Request, Response}
+  require Finch.Pool
 
   use Supervisor
 
@@ -89,6 +90,14 @@ defmodule Finch do
   @type stream(acc) ::
           ({:status, integer} | {:headers, Mint.Types.headers()} | {:data, binary}, acc -> acc)
 
+  @typedoc """
+  The reference used to identify an async request sent using `async_request/3`.
+
+  This reference can be used to identify process messages associated with the request,
+  but its shape should not be relied on and is subject to change.
+  """
+  @opaque request_ref() :: Finch.Pool.request_ref()
+
   @doc """
   Start an instance of Finch.
 
@@ -115,7 +124,6 @@ defmodule Finch do
       registry_name: name,
       manager_name: manager_name(name),
       supervisor_name: pool_supervisor_name(name),
-      task_supervisor_name: task_supervisor_name(name),
       default_pool_config: default_pool_config,
       pools: pools
     }
@@ -134,7 +142,6 @@ defmodule Finch do
   def init(config) do
     children = [
       {DynamicSupervisor, name: config.supervisor_name, strategy: :one_for_one},
-      {Task.Supervisor, name: config.task_supervisor_name},
       {Registry, [keys: :duplicate, name: config.registry_name, meta: [config: config]]},
       {PoolManager, config}
     ]
@@ -225,7 +232,6 @@ defmodule Finch do
   defp supervisor_name(name), do: :"#{name}.Supervisor"
   defp manager_name(name), do: :"#{name}.PoolManager"
   defp pool_supervisor_name(name), do: :"#{name}.PoolSupervisor"
-  defp task_supervisor_name(name), do: :"#{name}.TaskSupervisor"
 
   defmacrop request_span(request, name, do: block) do
     quote do
@@ -283,18 +289,8 @@ defmodule Finch do
   end
 
   defp __stream__(%Request{} = req, name, acc, fun, opts) when is_function(fun, 2) do
-    shp = build_shp(req)
-    {pool, pool_mod} = PoolManager.get_pool(name, shp)
+    {pool, pool_mod} = get_pool(req, name)
     pool_mod.request(pool, req, acc, fun, opts)
-  end
-
-  defp build_shp(%Request{scheme: scheme, unix_socket: unix_socket})
-       when is_binary(unix_socket) do
-    {scheme, {:local, unix_socket}, 0}
-  end
-
-  defp build_shp(%Request{scheme: scheme, host: host, port: port}) do
-    {scheme, host, port}
   end
 
   @doc """
@@ -363,8 +359,6 @@ defmodule Finch do
     end
   end
 
-  @opaque request_ref() :: {pid(), reference()}
-
   @doc """
   Sends an HTTP request asynchronously, returning a request reference.
 
@@ -373,43 +367,25 @@ defmodule Finch do
   """
   @spec async_request(Request.t(), name(), keyword()) :: request_ref()
   def async_request(%Request{} = req, name, opts \\ []) do
-    # TODO: get supervisor from config when this moves into the adapter
-    sup = task_supervisor_name(name)
-
-    %{pid: pid, ref: ref} =
-      Task.Supervisor.async_nolink(sup, __MODULE__, :__async_request__, [self(), req, name, opts])
-
-    send(pid, {:ref, ref})
-
-    {pid, ref}
-  end
-
-  @doc false
-  def __async_request__(owner, req, name, opts) do
-    ref = receive do: ({:ref, ref} -> ref)
-    request_ref = {self(), ref}
-
-    case stream(req, name, {owner, request_ref}, &send_chunk/2, opts) do
-      {:ok, _} -> send(owner, {request_ref, :done})
-      {:error, exception} -> send(owner, {request_ref, {:error, exception}})
-    end
-  end
-
-  defp send_chunk(chunk, {owner, request_ref}) do
-    send(owner, {request_ref, chunk})
-    {owner, request_ref}
+    {pool, pool_mod} = get_pool(req, name)
+    pool_mod.async_request(pool, req, opts)
   end
 
   @doc """
   Cancels a request sent with `async_request/3`.
   """
-  @spec cancel_async_request(request_ref(), name()) :: :ok | {:error, :not_found}
-  def cancel_async_request(request_ref, name) do
-    # TODO: get supervisor from config when this moves into the adapter
-    sup = task_supervisor_name(name)
+  @spec cancel_async_request(request_ref()) :: :ok
+  def cancel_async_request(request_ref) when Finch.Pool.is_request_ref(request_ref) do
+    {_ref, _pool, pool_mod, _state} = request_ref
+    pool_mod.cancel_async_request(request_ref)
+  end
 
-    {pid, ref} = request_ref
-    Process.demonitor(ref, [:flush])
-    Task.Supervisor.terminate_child(sup, pid)
+  defp get_pool(%Request{scheme: scheme, unix_socket: unix_socket}, name)
+       when is_binary(unix_socket) do
+    PoolManager.get_pool(name, {scheme, {:local, unix_socket}, 0})
+  end
+
+  defp get_pool(%Request{scheme: scheme, host: host, port: port}, name) do
+    PoolManager.get_pool(name, {scheme, host, port})
   end
 end
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -103,7 +103,9 @@ defmodule Finch.HTTP1.Pool do
   end
 
   @impl Finch.Pool
-  def cancel_async_request(_request_ref) do
+  def cancel_async_request({_, _, _, pid}) do
+    Process.exit(pid, :shutdown)
+    :ok
   end
 
   @impl NimblePool
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -74,6 +74,38 @@ defmodule Finch.HTTP1.Pool do
     end
   end
 
+  @impl Finch.Pool
+  def async_request(pool, req, opts) do
+    owner = self()
+
+    pid =
+      spawn(fn ->
+        request_ref = receive_next_within(10)
+        request(pool, req, {owner, request_ref}, &send_async_response/2, opts)
+        send(owner, {request_ref, :done})
+      end)
+
+    request_ref = {make_ref(), pool, __MODULE__, pid}
+    send(pid, request_ref)
+  end
+
+  defp receive_next_within(timeout) do
+    receive do
+      value -> value
+    after
+      timeout -> raise "no message received within #{timeout}"
+    end
+  end
+
+  defp send_async_response(response, {owner, request_ref}) do
+    send(owner, {request_ref, response})
+    {owner, request_ref}
+  end
+
+  @impl Finch.Pool
+  def cancel_async_request(_request_ref) do
+  end
+
   @impl NimblePool
   def init_pool({registry, shp, opts}) do
     # Register our pool with our module name as the key. This allows the caller
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -80,16 +80,20 @@ defmodule Finch.HTTP1.Pool do
 
     pid =
       spawn(fn ->
-        request_ref = receive_next_within(10)
-        request(pool, req, {owner, request_ref}, &send_async_response/2, opts)
-        send(owner, {request_ref, :done})
+        monitor = Process.monitor(owner)
+        request_ref = receive_next_within!(10)
+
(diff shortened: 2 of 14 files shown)


wojtekmach:
Looking forward to this!

If you don't mind me reqing this thread, here's a proof of concept of using this branch: (thanks to the enhanced `:finch_request` option also done by @zachallaun :D)

```elixir
Mix.install([
  :req,
  {:finch, github: "zachallaun/finch", branch: "async", override: true},
  :bandit
])

defmodule MyPlug do
  use Plug.Router

  plug(:match)
  plug(:dispatch)

  get "/redirect" do
    to = "/stream"

    conn
    |> Plug.Conn.put_resp_header("location", to)
    |> Plug.Conn.send_resp(302, "redirecting to #{to}")
  end

  get "/stream" do
    conn = Plug.Conn.send_chunked(conn, 200)

    Enum.reduce_while(~w(a b c d), conn, fn chunk, conn ->
      IO.inspect({:send, chunk})

      case Plug.Conn.chunk(conn, chunk) do
        {:ok, conn} ->
          {:cont, conn}

        {:error, :closed} ->
          {:halt, conn}
      end
    end)
  end
end

fun = fn request, finch_request, finch_name, finch_options ->
  ref = Finch.async_request(finch_request, finch_name, finch_options)

  receive do
    {^ref, {:status, status}} ->
      receive do
        {^ref, {:headers, headers}} ->
          body =
            Stream.unfold(ref, fn ref ->
              receive do
                {^ref, {:data, data}} ->
                  IO.inspect({:recv, data})
                  {data, ref}

                {^ref, :done} ->
                  nil
              end
            end)

          {request, Req.Response.new(status: status, headers: headers, body: body)}
      end
  end
end

{:ok, _} = Bandit.start_link(plug: MyPlug, scheme: :http, port: 4000)

stream = Req.get!("http://localhost:4000/redirect", finch_request: fun).body
IO.inspect({:take, Enum.take(stream, 1)})
Process.sleep(1000)
IO.inspect({:take, Enum.take(stream, 10)})
```

Outputs:

```
11:37:41.997 [info] Running MyPlug with Bandit 0.7.7 at 0.0.0.0:4000 (http)

11:37:42.057 [debug] follow_redirects: redirecting to /stream
{:send, "a"}
{:send, "b"}
{:send, "c"}
{:recv, "a"}
{:send, "d"}
{:take, ["a"]}
{:recv, "b"}
{:recv, "c"}
{:recv, "d"}
{:take, ["b", "c", "d"]}
```

I'm curious if it's OK for Req to consume :status and :headers bits like that so it can return a response struct (which we can then pass through response steps)? Or that's undesirable/unsafe? I think I read somewhere that a server may return additional headers down the road. Not sure if that's true or how popular that is but it would break the model above.

Otherwise we'd go with some kind of Req.stream which would NOT use response steps and that'd be OK too.

wojtekmach:
Yeah servers can send [Trailer](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer) headers. :|

zachallaun:
> If you don't mind me reqing this thread, here's a proof of concept of using this branch

I don’t mind at all! While I hope that others will find good use for this feature, integration with Req is my own primary use-case :)

So: yes, servers can sent trailing headers, and I also think the Mint docs say that you can get multiple headers chunks, so this is theoretically a valid series:

```
:status
:headers
:headers
:data
:data
:data
:headers
:done
```

That said… I think there’s a way for Req to take this into consideration to allow processing in response steps. (Not doing so would be, imo, a big missed opportunity for Req!) I’m frankly guessing, but I don’t think trailing headers are all that common…?

What if stream steps get a stream of PartialResponse structs? Req consumes the stream until it receives the first :data, then starts emitting partial responses containing status, all headers, and that data chunk? If more headers show up, the next partial response includes them. 

(Also, I’m happy discussing wherever you think is best, but maybe it makes sense to move this to the streaming issue in Req since I know there were other folks interested.)

zachallaun:
(Unrelated to the discussion with Wojtek above)

If anyone has a script that can be used for benchmarking Finch, I'd love to see it! As mentioned in the OP, I'd like to make sure that this branch is benchmarked against `main` to ensure no performance regressions for both HTTP1 and 2.

wojtekmach:
We can continue the conversation elsewhere, good call! https://github.com/wojtekmach/req/issues/82

zachallaun:
There are still a few things to address, I'd appreciate review at this point! Please see the OP for remaining TODOs and some additional notes.

zachallaun:
Wrote up a basic benchmark, would appreciate someone else taking a look at some point.

[Benchmark](https://github.com/zachallaun/finch/blob/async-bench/bench_finch.exs)

```
❯ elixir bench_finch.exs 
==> finch
Compiling 5 files (.ex)
Operating System: Linux
CPU Information: Intel(R) Core(TM) i5-4690K CPU @ 3.50GHz
Number of Available Cores: 4
Available memory: 15.59 GB
Elixir 1.14.5
Erlang 25.3.2

Benchmark suite executing with the following configuration:
warmup: 2 s
time: 5 s
memory time: 0 ns
reduction time: 0 ns
parallel: 1
inputs: http1, http2
Estimated total run time: 14 s

Benchmarking Finch.request!/3 with input http1 ...

15:29:14.187 [info] Running PlugServer with Bandit 1.0.0-pre.5 at 127.0.0.1:4000 (http)
Benchmarking Finch.request!/3 with input http2 ...

15:29:21.278 [info] Running PlugServer with Bandit 1.0.0-pre.5 at 127.0.0.1:4000 (https)

##### With input http1 #####
Name                               ips        average  deviation         median         99th %
Finch.request!/3                5.50 K      181.82 μs    ±32.06%      170.80 μs      379.29 μs
Finch.request!/3 (saved)        5.48 K      182.50 μs    ±29.53%      172.30 μs      305.90 μs

Comparison: 
Finch.request!/3                5.50 K
Finch.request!/3 (saved)        5.48 K - 1.00x slower +0.68 μs

##### With input http2 #####
Name                               ips        average  deviation         median         99th %
Finch.request!/3 (saved)        2.62 K      382.01 μs    ±23.88%      348.60 μs      700.08 μs
Finch.request!/3                2.60 K      385.05 μs    ±20.14%      356.60 μs      661.72 μs

Comparison: 
Finch.request!/3 (saved)        2.62 K
Finch.request!/3                2.60 K - 1.01x slower +3.03 μs
```

To reproduce:

```sh
# in your local finch project
curl -o bench_finch.exs https://github.com/zachallaun/finch/blob/async-bench/bench_finch.exs

git checkout main
elixir bench_finch.exs

git checkout async
elixir bench_finch.exs

# set SAVE env var to overwrite saved results
SAVE=1 elixir bench_finch.exs
```

zachallaun:
I started work on telemetry but hit a small roadblock. It's somewhat easy to move the `:send` measurement into the HTTP/2 pool, but the `:recv` is trickier because it currently includes the status and headers as telemetry metadata. The pool doesn't track any of that right now -- it just sends it along to the calling process, which collects it in a `receive` loop.

Do we want to save `:status` and `:headers` in the `%RequestStream{}` so that the pool has them available as metadata to emit telemetry events?

sneako (commented):
This is some great work @zachallaun!

I ran these changes for both h1 and h2 through a benchmarking setup that I have used for Finch in the past and the results agreed with your benchmarks, there do not seem to be any performance regressions with these changes.

RE: H2 telemetry, storing the status and headers in the RequestStream sound fine to me.

lib/finch.ex
```diff
@@ -351,12 +350,84 @@ defmodule Finch do
...
+  ## Responses
+
+  Response information is sent to the calling process as it is received
+  in `{ref, response}` tuples. Responses include:
+
+    * `{:status, status}` - HTTP response status
+    * `{:headers, headers}` - HTTP response headers
+    * `{:data, data}` - section of the HTTP response body
+    * `{:error, exception}` - an error occurred during the request
+    * `:done` - request has completed successfully
```
josevalim, line 374: Should we support trailing headers just in case?

```suggestion
    * `{:done, trailing_headers}` - request has completed successfully
```

?

josevalim, line 374: Maybe we should have trailing headers as part of `:done`?

```suggestion
    * `{:done, trailing_headers}` - request has completed successfully
```

Thoughts?

lib/finch.ex
```diff
@@ -351,12 +350,84 @@ defmodule Finch do
...
+    {pool, pool_mod} = get_pool(req, name)
+    pool_mod.async_request(pool, req, opts)
+  end
+
+  @doc """
+  Cancels a request sent with `async_request/3`.
+  """
+  @spec cancel_async_request(request_ref()) :: :ok
+  def cancel_async_request(request_ref) when Finch.Pool.is_request_ref(request_ref) do
+    {_ref, _pool, pool_mod, _state} = request_ref
```
josevalim, line 421: I am worried that, if we have to keep any state, we now have to pass the copy of this state on every message. Maybe `async_request` should return `{stream_ref, cancel_ref}`?

lib/finch/http1/pool.ex
```diff
@@ -74,6 +74,56 @@ defmodule Finch.HTTP1.Pool do
     end
   end
 
+  @impl Finch.Pool
+  def async_request(pool, req, opts) do
+    owner = self()
+
+    pid =
+      spawn(fn ->
```
josevalim, line 82: Maybe we should link?

```suggestion
      spawn_link(fn ->
```

This will make sure any unexpected crashes will propagate up.

lib/finch/http1/pool.ex
```diff
@@ -74,6 +74,56 @@ defmodule Finch.HTTP1.Pool do
...
+        monitor = Process.monitor(owner)
+        request_ref = receive_next_within!(10)
```
josevalim, line 84: We would definitely want this value to be much higher. Otherwise a temporarily slow system will start seeing crashes.

```suggestion
        request_ref = receive_next_within!(5_000)
```

lib/finch.ex
josevalim: Unless this version is much harder to implement.

lib/finch/http1/pool.ex
josevalim: If we link, there may be no need to monitor, as this process we crash. Alternatively we link, trap exits, and monitor.

lib/finch.ex
zachallaun: I don’t think it would be much harder. Right now it is 1-1 with the responses that Mint.stream returns, which doesn’t differentiate in the response shape between normal and trailing headers. 

So the thing to consider is that we’ll receive both trailing headers and a `:done` from Mint. Do we want to cache trailing headers and emit them if present when we receive `:done`, or would we want to immediately omit `:done` with those headers immediately and ignore the done response when we get it?

In my opinion, it’s not worth treating them specially — we just document (like Mint does) that an additional :headers received at the end of the response just prior to :done are trailing headers.

lib/finch.ex
zachallaun: I don’t feel strongly about this. Right now, HTTP1 uses the state to store the pid of the spawned process wrapping the blocking `request` call, and HTTP2 doesn’t use the state at all (field is `nil`). 

So HTTP1 needs:
- pool mod to know it’s HTTP1
- PID of spawned process (will be unique per request, so we don’t even need a reference here)

HTTP2 needs:
- pool mod to know it’s HTTP2
- pool pid to know which pool process is handling the request/response
- reference to differentiate between requests

So if we wanted to simplify this value, it could be:

```elixir
module Finch.Pool do
  @type cancel_ref :: term()
 
  @callback async_request(…) :: cancel_ref()
  @callback cancel_async_request(cancel_ref()) :: :ok

  ...
end

defmodule Finch do
  @opaque request_ref :: {pool_mod :: module, Finch.Pool.cancel_ref()}
end

# HTTP1
{__MODULE__, pid}

# HTTP2
{__MODULE__, {pool_pid, make_ref()}}
```

And let each pool keep only what’s needed to cancel the request. This would, I think, better communicate to authors in the future that this shouldn’t be an arbitrary bunch of state, but only what’s needed to cancel an in-flight request.

lib/finch/http1/pool.ex
zachallaun: Yep, that would definitely simplify things. I was thinking that we wouldn’t want internal crashes to take down the caller, only the caller to take down the request, but the request crashing already means the caller isn’t going to receive messages it’s probably expecting, so it’s probably better to just crash.

lib/finch.ex
josevalim: Agreed then. Let’s leave it as is!

zachallaun:
Thank you both for the review!

@sneako How would you like to handle #227? Both it and this PR touch the H2 pool and tests and will need to resolve conflicts. If you anticipate merging #227 first, I’m happy to do the work of rebasing this PR!

lib/finch.ex
josevalim: > So if we wanted to simplify this value, it could be:

Yes, this would be nice (and enough to keep me quiet). :)

zachallaun: I never want you to stay quiet! 😄

zachallaun:
Regarding H2 telemetry, here’s what I’m thinking:

- `:connect` start/stop - already being sent from within the pool process
- `:send` start/stop - move into the pool process, save the start_time on the request state and emit stop when we reply to sync requests
- `:recv` start/stop - pretty easy, we’ll store the status/headers as suggested and emit stop when we pop the request out of the state. This should handle both regular stop because we received `:done` and cancelled requests. 
- `:recv` exception is a little trickier for the case where the pool crashes. I suppose we should trap exits in the pool and then we can emit this for any in-process requests before shutting down? @josevalim, any suggestions/thoughts here?

lib/finch/http1/pool.ex
zachallaun: @josevalim See 9ae43ef7f71aef81583e2ff1ef74512613ce788d -- do you have any insight into why the `spawn_link` process doesn't immediately terminate when the caller does? Specifically, [this test](https://github.com/sneako/finch/pull/228/files#diff-0ce8331b6dbdaa98b49a795e62340ca7bb472824c919d4fafb94936967688745R102) was failing when I stripped out the monitoring code. I also tried adding `Process.flag(:trap_exit, true)` and then changing `process_down?(monitor)` to a `should_exit()` that looks for an `{:EXIT, _, _}` message, but didn't seem to receive one.

lib/finch.ex
zachallaun: Done in f89106b926a3490f226466ee1d6c199236c90374, not quite exactly as above, but definitely simpler and allowed getting rid of `receive_next_within!` completely.

lib/finch/http1/pool.ex
josevalim: In this code:

```elixir
        spawn(fn ->
          ref =
            Finch.build(:get, url <> "/stream/5/500")
            |> Finch.async_request(finch_name)

          send(outer, ref)
        end)
```

the spawned process immediately exits with reason `:normal` after `send/2`, which does not lead to any broken link. So by the time you send it a exit signal, it is already gone. Adding a `Process.sleep(:infinity)` should fix it.

It it up to you to decide if we still need monitoring. If you start something to stream asynchronously and you never consume it, what does it mean?

josevalim:
Shall we add the remaining bits about H2 telemetry in another PR/discussion? I am actually not sure how much I can help here. It may be worth pinging `@whatyouhide` (quoted so it doesn't ping him yet).

lib/finch/http1/pool.ex
zachallaun: > the spawned process immediately exits with reason `:normal` after `send/2`, which does not lead to any broken link. So by the time you send it a exit signal, it is already gone. Adding a `Process.sleep(:infinity)` should fix it.

Oh, duh! Thank you. 

I’ll fix that and see how it behaves without monitoring and decide from there.

zachallaun:
> I am actually not sure how much I can help here.

You’ve been a great help so far! I’ll leave it up to @sneako as to whether we group H2 telemetry into this or another PR, but please feel free to unsub from this PR to reduce chatter for now and we can ping you if needed!

lib/finch/http1/pool.ex
zachallaun: Split the test to explicitly test both normal and abnormal caller exits. Decided to keep the monitoring around because I think canceling the request if there is no caller to receive it is the right thing to do. I'll document this behavior in `async_request/3`, so that folks know that the caller must stick around for the request to continue. I can imagine some potentially useful patterns around this behavior, e.g.

```elixir
Task.async(fn ->
  some_streamed_request()
  |> Enum.find(fn
    {:data, data} -> some_chunk_im_looking_for?(data)
    _ -> false
  end)
end)
```

sneako:
> Thank you both for the review!
> 
> @sneako How would you like to handle #227? Both it and this PR touch the H2 pool and tests and will need to resolve conflicts. If you anticipate merging #227 first, I’m happy to do the work of rebasing this PR!

Don't worry about #227, let's merge this PR first and then I can rebase that much smaller PR if necessary. 

I'm ok with continuing h2 telemetry updates in a subsequent PR, there is already a lot going on here and I think consumers like Req can already start playing with the api without all of the telemetry in place.

zachallaun:
> I'm ok with continuing h2 telemetry updates in a subsequent PR, there is already a lot going on here and I think consumers like Req can already start playing with the api without all of the telemetry in place.

Sounds good to me!

For now, should we just add a note to the top of the `Finch.Telemetry` moduledoc? Thinking something like:

> Note that for HTTP2 requests issued with `Finch.async_request/3`, most telemetry events are not currently emitted. This will be addressed in a later release.

zachallaun:
@sneako Do you perchance know what's going on with [this ranch error](https://github.com/sneako/finch/actions/runs/5236837833/jobs/9454854009#step:12:26)? Only on OTP 26.

zachallaun:
Hmm, it's curious that the [error is still there](https://github.com/sneako/finch/actions/runs/5237209681/jobs/9455318948?pr=228#step:12:12), but the tests are still "passing" in CI despite skipping a number of tests -- it's showing 86 tests run, but there are 90.

josevalim:
Agreed. The test suite should be failing. A bug report/PR is welcome!

sneako:
Opened https://github.com/elixir-lang/elixir/issues/12650

zachallaun:
@sneako So the problem is with [this line](https://github.com/zachallaun/finch/blob/async/test/support/https1_server.ex#L17). Maybe this is something that's changed in OTP26 (there were [changes to SSL](https://www.erlang.org/blog/otp-26-highlights/#ssl-safer-defaults)), but this needs to be a list of strings. Changing it to `alpn_preferred_protocols: ["h2"]` gets rid of the error, but then [this test fails](https://github.com/zachallaun/finch/blob/async/test/finch/http1/integration_test.exs#L19).

I admit that this is somewhat beyond my understanding. If you have any suggestions about what to do with this test, I'd appreciate it!

sneako:
Didn't have too much time last night to dig in, but here is the commit that changed this behaviour in OTP: https://github.com/erlang/otp/commit/cd00692e18cc88c369e0a4fdaa124bdea7d67252#diff-c6e4e01fbddfaa6057fc43144940a337eb3db17891c62cce9ad2123e0107f776R2024

It does state: "This commit have changed error messages for a couple of error cases, and is also more restrict with checking the options."

I think they do intend for `:undefined` to be a valid value since that is [still the default](https://github.com/erlang/otp/blob/master/lib/ssl/src/ssl.erl#L2090), but I'll try to dig a bit more before opening an issue with them

zachallaun:
Thanks for looking into it!

Given that this is somewhat tangential to this PR, is it worth opening a new issue/PR to address it later?

sneako:
Sure! Let's just remove OTP 26 from the test matrix for now

lib/finch/http1/pool.ex
```diff
@@ -74,6 +74,48 @@ defmodule Finch.HTTP1.Pool do
...
+    if process_down?(monitor) do
+      exit(:shutdown)
+    end
+
+    send(owner, {request_ref, response})
+    {owner, monitor, request_ref}
+  end
+
+  defp process_down?(monitor) do
+    receive do
```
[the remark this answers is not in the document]
sneako: ah and of course, you have a monitor here, not a pid :)

hkrutzer:
Thanks @zachallaun this is a great improvement for the entire ecosystem.

wojtekmach:
Thanks again for everyone involved!

I was playing with this API in a test:

```elixir
Mix.install([
  {:finch, github: "sneako/finch"},
  :bypass
])

ExUnit.start()

defmodule MyApp.Test do
  use ExUnit.Case

  test "it works" do
    bypass = Bypass.open()

    Bypass.expect(bypass, "GET", "/", fn conn ->
      conn = Plug.Conn.send_chunked(conn, 200)
      stream = Stream.map(1..5, &to_string/1)

      Enum.reduce_while(stream, conn, fn chunk, conn ->
        Process.sleep(100)
        IO.inspect(chunk, label: :send)

        case Plug.Conn.chunk(conn, chunk) do
          {:ok, conn} ->
            {:cont, conn}

          {:error, :closed} ->
            {:halt, conn}
        end
      end)
    end)

    start_supervised!({Finch, name: MyFinch})

    ref =
      Finch.build(:get, "http://0.0.0.0:#{bypass.port}")
      |> Finch.async_request(MyFinch)

    receive do
      {^ref, chunk} -> dbg(chunk)
    end
  end
end
```

And I got this:

```
[stream.exs:39: MyApp.Test."test it works"/1]
chunk #=> {:status, 200}



  1) test it works (MyApp.Test)
     stream.exs:11
     ** (exit) shutdown


Finished in 0.05 seconds (0.01s on load, 0.00s async, 0.04s sync)
1 test, 1 failure
```

I think the test process exited while the request was still going but per docs:

> If the calling process exits before the request has completed, the request will be canceled.

zachallaun:
Thanks for reporting! I had this issue at one point while developing the
feature but thought it was accounted for. Apparently not! I’ll look into
this more tonight.

zachallaun:
Okay, so after a bit more testing, I'm not sure this has anything to do with `async_request` but was behavior that was already present when using Bypass previously. Here's a replication of the same behavior using Finch 0.16.0:

```elixir
Mix.install([
  {:finch, "0.16.0"},
  :bypass
])

ExUnit.start()

defmodule MyApp.Test do
  use ExUnit.Case

  test "it works" do
    bypass = Bypass.open()

    Bypass.expect(bypass, "GET", "/", fn conn ->
      conn = Plug.Conn.send_chunked(conn, 200)
      stream = Stream.map(1..5, &to_string/1)

      Enum.reduce_while(stream, conn, fn chunk, conn ->
        Process.sleep(100)
        IO.inspect(chunk, label: :send)

        case Plug.Conn.chunk(conn, chunk) do
          {:ok, conn} ->
            {:cont, conn}

          {:error, :closed} ->
            {:halt, conn}
        end
      end)
    end)

    start_supervised!({Finch, name: MyFinch})

    outer = self()

    spawn_link(fn ->
      Finch.build(:get, "http://0.0.0.0:#{bypass.port}")
      |> Finch.stream(MyFinch, nil, fn chunk, _ ->
        send(outer, chunk)
      end)
    end)

    receive do
      chunk -> dbg(chunk)
    end
  end
end
```

Running this:

```
❯ elixir scratch.exs
[scratch.exs:44: MyApp.Test."test it works"/1]
chunk #=> {:status, 200}



  1) test it works (MyApp.Test)
     scratch.exs:11
     ** (exit) shutdown


Finished in 0.3 seconds (0.1s on load, 0.00s async, 0.1s sync)
1 test, 1 failure
```

I don't fully understand Bypass, but somehow the underlying test server is receiving an exit message when the test exits, which then triggers Bypass to [re-raise and exit](https://github.com/PSPDFKit-labs/bypass/blob/v2.1.0/lib/bypass.ex#L132) when it's verifying expectations in a test `on_exit` callback.

We can verify this by copying what happens in `Bypass.open()` and setting our own `on_exit`:

```elixir
Mix.install([
  {:finch, "0.16.0"},
  :bypass
])

ExUnit.start()

defmodule MyApp.Test do
  use ExUnit.Case

  test "it works" do
    {:ok, pid} = DynamicSupervisor.start_child(Bypass.Supervisor, Bypass.Instance.child_spec([]))
    port = Bypass.Instance.call(pid, :port)
    bypass = %Bypass{pid: pid, port: port}

    on_exit(fn ->
      dbg(Bypass.Instance.call(bypass.pid, :on_exit))
    end)

    Bypass.expect(bypass, "GET", "/", fn conn ->
      conn = Plug.Conn.send_chunked(conn, 200)
      stream = Stream.map(1..5, &to_string/1)

      Enum.reduce_while(stream, conn, fn chunk, conn ->
        Process.sleep(100)
        IO.inspect(chunk, label: :send)

        case Plug.Conn.chunk(conn, chunk) do
          {:ok, conn} ->
            {:cont, conn}

          {:error, :closed} ->
            {:halt, conn}
        end
      end)
    end)

    start_supervised!({Finch, name: MyFinch})

    outer = self()

    spawn_link(fn ->
      Finch.build(:get, "http://0.0.0.0:#{bypass.port}")
      |> Finch.stream(MyFinch, nil, fn chunk, _ ->
        send(outer, chunk)
      end)
    end)

    receive do
      chunk -> dbg(chunk)
    end
  end
end
```

```
❯ elixir scratch.exs
[scratch.exs:50: MyApp.Test."test it works"/1]
chunk #=> {:status, 200}

[scratch.exs:17: MyApp.Test."test it works"/1]
Bypass.Instance.call(bypass.pid, :on_exit) #=> {:exit, {:exit, :shutdown, []}}

.
Finished in 0.1 seconds (0.08s on load, 0.00s async, 0.1s sync)
1 test, 0 failures
```

I'm not 100% sure what leads to this, but I think the Plug.Cowboy test server being shut down early when the test exits (the Bypass instance monitors it [here](https://github.com/PSPDFKit-labs/bypass/blob/v2.1.0/lib/bypass/instance.ex#L354) and saves the exit as a result [here](https://github.com/PSPDFKit-labs/bypass/blob/v2.1.0/lib/bypass/instance.ex#L61)).

Overall, I'm not sure that there's anything to be done in Finch. This sort of feels like a design flaw of Bypass with regards to streaming responses that may not complete.

wojtekmach:
Oh wow, excellent research. Sorry about that!

zachallaun:
Nothing to apologize for! If you run into anything else that seems weird while testing it out, please ping me.

Issue #229, opened 2023-06-13
Finch.async_request/3 does not emit all of the expected Telemetry events


sneako:
See https://github.com/sneako/finch/pull/228#issuecomment-1586151203 for additional context

PR #231, opened 2023-06-13
Emit `:send` and `:recv` telemetry for HTTP2 async requests

zachallaun:
Addresses #229.

### DIFF
--- a/lib/finch/http2/pool.ex
+++ b/lib/finch/http2/pool.ex
@@ -33,9 +33,7 @@ defmodule Finch.HTTP2.Pool do
     timeout = opts[:receive_timeout]
     request_ref = make_request_ref(pool)
 
-    metadata = %{request: request}
-
-    with :ok <- :gen_statem.call(pool, {:request, request_ref, request, opts}) do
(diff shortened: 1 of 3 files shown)


lib/finch/http2/pool.ex
```diff
@@ -33,37 +33,20 @@ defmodule Finch.HTTP2.Pool do
     timeout = opts[:receive_timeout]
     request_ref = make_request_ref(pool)
 
-    metadata = %{request: request}
-    start_time = Telemetry.start(:send, metadata)
-
-    with :ok <- :gen_statem.call(pool, {:request, request_ref, request, opts}) do
-      Telemetry.stop(:send, start_time, metadata)
+    with {:ok, recv_start} <- :gen_statem.call(pool, {:request, request_ref, request, opts}) do
```
zachallaun, line 36: I don't love this, but I'm not sure there's a good way around it, since we need the `:recv` start time to emit an exception if the given `request_fun` raises in the response receive loop.

sneako:
Thanks again, @zachallaun !

PR #232, opened 2023-06-27
Fix calls to deprecated Logger.warn/2

solar05:
I'm not sure if this is the right way. Maybe it's also possible to bump the minimum Elixir version to 1.11? But I guess that would be a breaking change :/

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         include:
           - pair:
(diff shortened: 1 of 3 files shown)


wojtekmach:
+1 for requiring Elixir 1.11.

wojtekmach:
If we merge this as is, we should update Elixir requirement in mix.exs.

solar05:
Done, also bump elixir CI version 👌

sneako:
Thank you!

PR #238, opened 2023-07-25
Do not use streams with async_request

### LINKED ISSUE #235: Do not use Finch.async_request with Stream
I can see that you've added `Finch.async_request` and doc even states this as a way to go to use it with `Stream.resource`, **but**

# The biggest problem

It doesn't lazily read the data from socket on demand, it just sends whole response body into message queue which can just overflow the server. Any huge (let's say 1TB) response will just shut down any program using this approach. And people usually use `Stream` for these kind of problems (requesting huge files). So this is just misleading and has nothing to do with stream, please remove this from your documentation

josevalim:
Closes #235.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -381,6 +381,10 @@ defmodule Finch do
   @doc """
   Sends an HTTP request asynchronously, returning a request reference.
 
+  This function works as a firehose and it will send messages to the
+  current process as the HTTP response arrives. If you want more control
+  over how messages are streamed, see `stream/5`.
+
   If the request is sent using HTTP1, an extra process is spawned to
   consume messages from the underlying socket. If you wish to maximize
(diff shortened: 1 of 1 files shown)


zachallaun:
Responding to your commend [here](https://github.com/sneako/finch/issues/235#issuecomment-1647615456):

> We should mention that `async_request` works as a firehose, it will immediately send messages as they come to the current process. I think discouraging its usage with `Stream` is fine. And we should mention `Finch.stream` as an option to lazily stream.

Unless I'm mistaken, `Finch.stream` is not an option to lazily stream, at least using HTTP2. Messages are consumed eagerly when received by the pool process and awaited in the [response waiting loop](https://github.com/sneako/finch/blob/main/lib/finch/http2/pool.ex#L80), which is where the function passed to `Finch.stream` runs.

PR #240, opened 2023-08-23
Update HTTP/2 pool log level to warning for retried action

### LINKED ISSUE #214: Feature Request: Adjust HTTP/2 log levels to reduce extra noise
Several events that happen down in the HTTP2 pool call `Logger.error` when they occur, including times when they will be automatically retried.

Most major exception trackers (Sentry, etc) have mechanism to look for log level or a particular level or higher. Customarily, this level defaults to `:error`. 

I'd love to be able to cut down the log level inside Finch, so these log message won't interfere with our error tracking.

HarshBalyan:
closes #214


sneako:
Thank you!

PR #241, opened 2023-08-28
Set MIX_ENV=test on CI

wojtekmach:
We have `mix deps.compile` and `mix compile` steps and before this patch
they were running in the default env, :dev, resulting in unnecessary
compilations.


sneako:
thank you!!

PR #242, opened 2023-08-29
Split trailers from headers

wojtekmach:
Ref https://github.com/wojtekmach/req/issues/230#issue-1869352131

I couldn't write an integration test for http2 given Plug has no support for trailers and it is not easy to write a fake server.

I verified it manually:

```
$ openssl req -newkey rsa:2048 -nodes -keyout server.key -x509 -days 365 -out server.crt
$ cat main.go
```

```go
package main

import (
  "log"
  "net/http"
)

func main() {
  srv := &http.Server{Addr: ":8080", Handler: http.HandlerFunc(handle)}
  log.Printf("Serving on https://0.0.0.0:8080")
  log.Fatal(srv.ListenAndServeTLS("server.crt", "server.key"))
}

func handle(w http.ResponseWriter, r *http.Request) {
  log.Printf("Got connection: %s", r.Proto)
  w.Header().Set("trailer", "x-foo")
  w.Write([]byte("Hello"))
  w.Header().Set("x-foo", "bar")
}
```

```
$ cat run.exs
```

```elixir
{:ok, _} =
  Finch.start_link(
    name: MyFinch,
    pools: %{
      default: [
        protocol: :http2,
        conn_opts: [
          transport_opts: [verify: :verify_none]
        ]
      ]
    }
  )

req = Finch.build(:get, "https://localhost:8080")
IO.inspect(Finch.request(req, MyFinch))
```

```
$ go run main.go
$ mix run run.exs
```

```elixir
{:ok,
 %Finch.Response{
   status: 200,
   body: "Hello",
   headers: [
     {"trailer", "x-foo"},
     {"content-type", "text/plain; charset=utf-8"},
     {"content-length", "5"},
     {"date", "Tue, 29 Aug 2023 09:37:39 GMT"}
   ],
   trailers: [{"x-foo", "bar"}]
 }}
```

### DIFF
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -312,7 +312,7 @@ defmodule Finch.Conn do
               mint,
               ref,
               timeout,
-              :trailers,
+              fields,
               resp_metadata
             )
 
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -298,9 +298,9 @@ defmodule Finch.Conn do
       {:headers, ^ref, value} ->
         resp_metadata =
           if fields == :headers do
-            %{resp_metadata | headers: value}
+            update_in(resp_metadata.headers, &(&1 ++ value))
           else
-            %{resp_metadata | trailers: value}
+            update_in(resp_metadata.trailers, &(&1 ++ value))
           end
 
         case fun.({fields, value}, acc) do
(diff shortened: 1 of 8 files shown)


lib/finch/http1/conn.ex
```diff
@@ -223,20 +283,27 @@ defmodule Finch.Conn do
...
             raise ArgumentError, "expected {:cont, acc} or {:halt, acc}, got: #{inspect(other)}"
         end
 
       {:headers, ^ref, value} ->
-        case fun.({:headers, value}, acc) do
+        resp_metadata =
+          if fields == :headers do
+            %{resp_metadata | headers: value}
+          else
+            %{resp_metadata | trailers: value}
```
sneako, line 303: isn't it possible that we could get multiple :headers or :trailers messages, meaning we need to concat `value` to the value we have in the accumulator?

sneako (approved):
Awesome, thanks @wojtekmach !

Issue #243, opened 2023-08-30
receive_timeout regarding stream/5 & async_request/3


jswanner:
Thanks for the amazing library!

`Finch.stream/5` seems to treat the `receive_timeout` option differently between v0.16.0 & main (9609c40), and I believe it stems from [823cc65](https://github.com/sneako/finch/commit/823cc65a25205fa1e9a9ce91c8acc58de65586fb).

On main, `Finch.stream/5` (and `Finch.async_request/3`) timeout after the `receive_timeout` threshold has elapsed, even if data are being received. While on v0.16.0, the connection times out after the `receive_timeout` threshold has elapsed since the last chunk of data (rather than from the start of the connection).

Maybe I'm wanting to do something that's not intended to be supported, which is to leave the streaming connection open as long as it's still receiving data (the API I'm calling should send pings every 5 seconds). I can work around the problem by setting a very high `receive_timeout` value and then reconnect on timeout, but it would be nice to somehow have the `receive_timeout` behavior of v0.16.0.

sneako:
Thank you for the kind words!

I see your point. 

I guess, the value that is passed to the recv calls should always be the `:receive_timeout` and something similar to the new behaviour that was introduced could be implemented as a new `:request_timeout` option for the overall duration of request.

jswanner:
> I guess, the value that is passed to the recv calls should always be the `:receive_timeout` and something similar to the new behaviour that was introduced could be implemented as a new `:request_timeout` option for the overall duration of request.

@sneako, what you proposed sounds good to me. I _thought_ I understood what needed to happen well enough to implement it myself, but turns out I _don't_ know what I'm doing. Mint seems to be pipelining requests, and the request `ref` is unexpectedly changing.

Would you like me to push a broken PR? Maybe you'll have some insight into what I'm doing wrong.

zachallaun:
> Would you like me to push a broken PR? Maybe you'll have some insight into what I'm doing wrong.

I'm not @sneako, but it seems totally reasonable to push it as a Draft PR!

jswanner:
Thanks for the prompt @zachallaun, I had been intending to make the PR, but had been focusing on other things.

sneako:
Closed via #244

jswanner:
Thank you, @sneako! ❤️

PR #244, opened 2023-09-01
Introduce :request_timeout option

jswanner:
Intended to timeout a request with a chunked response after the total elapsed time has surpassed the given value. See #243.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -399,8 +399,13 @@ defmodule Finch do
     * `:pool_timeout` - This timeout is applied when we check out a connection from the pool.
       Default value is `5_000`.
 
-    * `:receive_timeout` - The maximum time to wait for a response before returning an error.
+    * `:receive_timeout` - The maximum time to wait for each chunk to be received before returning an error.
       Default value is `15_000`.
+
+    * `:request_timeout` - The amount of time to wait for a complete response before returning an error.
+      This timeout only applies to HTTP/1, and its current implementation is a best effort timeout,
+      it does not guarantee the call will return precisely when the time has elapsed.
+      Default value is `:infinity`.
 
   """
   @spec request(Request.t(), name(), request_opts()) ::
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -93,9 +93,9 @@ defmodule Finch.Conn do
     end
   end
 
-  def request(%{mint: nil} = conn, _, _, _, _, _), do: {:error, conn, "Could not connect"}
+  def request(%{mint: nil} = conn, _, _, _, _, _, _), do: {:error, conn, "Could not connect"}
 
-  def request(conn, req, acc, fun, receive_timeout, idle_time) do
+  def request(conn, req, acc, fun, receive_timeout, request_timeout, idle_time) do
     full_path = Finch.Request.request_path(req)
 
     metadata = %{request: req}
@@ -113,11 +113,12 @@ defmodule Finch.Conn do
              stream_or_body(req.body)
            ) do
         {:ok, mint, ref} ->
-          case maybe_stream_request_body(mint, ref, req.body, receive_timeout) do
+          case maybe_stream_request_body(mint, ref, req.body) do
             {:ok, mint} ->
               Telemetry.stop(:send, start_time, metadata, extra_measurements)
               start_time = Telemetry.start(:recv, metadata, extra_measurements)
               resp_metadata = %{status: nil, headers: [], trailers: []}
+              timeouts = %{receive_timeout: receive_timeout, request_timeout: request_timeout}
 
               response =
                 receive_response(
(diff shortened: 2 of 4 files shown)


jswanner:
The test suite currently fails with this change with the following:

```
  1) test request/3 returns error when request times out for chunked response (FinchTest)
     test/finch_test.exs:503
     ** (CaseClauseError) no case clause matching: {:data, #Reference<0.1536038530.2835349507.48741>, "chunk-data"}
     code: |> Finch.request(finch_name, request_timeout: timeout * 10)
     stacktrace:
       (finch 0.16.0) lib/finch/http1/conn.ex:285: Finch.Conn.receive_response/9
       (finch 0.16.0) lib/finch/http1/conn.ex:122: Finch.Conn.request/7
       (finch 0.16.0) lib/finch/http1/pool.ex:47: anonymous fn/9 in Finch.HTTP1.Pool.request/5
       (nimble_pool 1.0.0) lib/nimble_pool.ex:349: NimblePool.checkout!/4
       (finch 0.16.0) lib/finch/http1/pool.ex:39: Finch.HTTP1.Pool.request/5
       (finch 0.16.0) lib/finch.ex:409: anonymous fn/4 in Finch.request/3
       (telemetry 1.0.0) /path/to/finch/deps/telemetry/src/telemetry.erl:293: :telemetry.span/3
       test/finch_test.exs:527: (test)
```

This `case` statement doesn't match because the `Reference` in `{:data, #Reference<>, "chunk-data"}` doesn't match `ref`:

https://github.com/sneako/finch/blob/d3c406bc29209cb9468c3a742d394960e11fedf1/lib/finch/http1/conn.ex#L285-L332

jswanner:
@sneako, I thought of a way to implement this that does work, but it does require spawning another process. I understand if that's an undesirable implementation detail.

sneako:
Thank you @jswanner but yes, spawning a process here is definitely not something we can add to the library, it will severely impact performance.

In order to have a very accurate and strict timeout, an additional process is probably necessary, but I was thinking in this case we can have a timeout with much looser guarantees that will still provide *some* value.

Basically this new timeout would be a value which will be decremented by the duration of each individual recv call. This should be "good enough" unless our calls to mint block indefinitely (shouldn't happen unless `:infinity` is passed in as the recv timeout) or if the Finch user is performing a lot of slow work in their stream function.

Does that make sense to you?

wojtekmach:
Sorry if this is silly but I always wondered if instead of doing the work in a separate process we could spawn a new process that would sleep for _timeout_ and kill the parent? Something like this:

```elixir
parent = self()

pid =
  spawn(fn ->
    Process.sleep(1000)
    Process.exit(parent, :timeout)
  end)

# do work...

Process.exit(pid, :normal)
```

Again this might be a dumb idea and so yeah I believe the best effort approach is warranted here. :)

@jswanner btw, I think I briefly saw you on the conference? If so I wish we had a chance to talk more about this!

jswanner:
> Basically this new timeout would be a value which will be decremented by the duration of each individual recv call. This should be "good enough" unless our calls to mint block indefinitely (shouldn't happen unless :infinity is passed in as the recv timeout) or if the Finch user is performing a lot of slow work in their stream function.
> 
> Does that make sense to you?

@sneako, it does make sense, and that's basically what I tried in the first commit of this PR. Unfortunately some weird stuff was happening in the test that I couldn't explain, it seems an additional request was queued and pipelined then data was received for a different request ref. 

@wojtekmach that's an interesting idea! I might have to play around with that just for my own curiosity. And yes, we did briefly chat at the very end of the conference.

jswanner:
@sneako, I've removed the spawning of a new process, and figured out what was going on. Let me know what you think.

lib/finch.ex
```diff
@@ -399,9 +399,12 @@ defmodule Finch do
     * `:pool_timeout` - This timeout is applied when we check out a connection from the pool.
       Default value is `5_000`.
 
-    * `:receive_timeout` - The maximum time to wait for a response before returning an error.
+    * `:receive_timeout` - The maximum time to wait for each chunk to be received before returning an error.
       Default value is `15_000`.
 
+    * `:request_timeout` - The maximum time to wait for a complete response before returning an error.
```
sneako, line 405: I think we should mention how this is a best effort timeout and that the current implementation does not guarantee that the call will return before this amount of time passes. We should also mention that this has only been implemented for HTTP/1.

sneako:
Thank you so much @jswanner and please accept my apologies for the delays, I've had some major life events happen recently but am finally feeling ready to get back on the horse. I would like to add a bit to the docs for the new option, but otherwise this looks great to me

jswanner:
@sneako, hopefully they were happy major life events and not sad ones.

Great points on the option's documentation. Let me know what you think of the changes I made.

sneako (approved):
Thank you @jswanner !! :heart:

Issue #199, opened 2022-06-08
Add Finch instance name to connect telemetry


aselder:
I'd like to be able to instrument connects from my finch instances. The various connect telemetry events don't expose the pool ID or Finch instance. I'd love to see this added to the meta data so that I can tag the stats I'm collecting by Finch instance and pool.

aselder:
Also, would love to see this added to the queue events as well.

sneako:
> Also, would love to see this added to the queue events as well.

Ideally we can add it to all events :pray:

sneako:
resolved in #252

PR #248, opened 2023-11-04
Implement pool telemetry

oliveigah:
Second proposal to solve https://github.com/sneako/finch/issues/183 

After gather some feedback, I settled down on these 2 structs for the metrics:

```elixir
   %Finch.HTTP1.PoolMetrics{
     pool_index: 1,
     pool_size: 50,
     available_connections: 40,
     in_use_connections: 10
   }
   
   %Finch.HTTP2.PoolMetrics{
      pool_index: 1,
      in_flight_requests: 4
   }
```

The `Finch.get_pool_status/2` interface also changed a bit. Now it returns a list of metrics struct, one for each pool started defined by the `count` option. Like this:

```elixir
Finch.get_pool_status(MyFinch, "https://somehost.com:4321")

{:ok,
 [
   %Finch.HTTP1.PoolMetrics{
     pool_index: 1,
     pool_size: 50,
     available_connections: 40,
     in_use_connections: 10
   },
   %Finch.HTTP1.PoolMetrics{
     pool_index: 2,
     pool_size: 50,
     available_connections: 40,
     in_use_connections: 10
   }
 ]}
```

I've also added some fields to the pool state (such as finch_name and pool_index) in order to enable future changes more easily.


lib/finch.ex
```diff
@@ -536,4 +542,9 @@ defmodule Finch do
   defp get_pool(%Request{scheme: scheme, host: host, port: port}, name) do
     PoolManager.get_pool(name, {scheme, host, port})
   end
+
+  def get_pool_status(finch_name, shp) do
```
sneako, line 546: Maybe we should also allow the user to pass the url as a string that we would parse into a shp just as a convenience function? We could mention in the docs that it is more efficient to pass the shp if that is of concern.

lib/finch/http1/pool.ex
```diff
@@ -116,34 +151,43 @@ defmodule Finch.HTTP1.Pool do
...
-  def init_worker({{scheme, host, port}, opts} = pool_state) do
+  def init_worker({_name, {scheme, host, port}, opts} = pool_state) do
     {:ok, Conn.new(scheme, host, port, opts, self()), pool_state}
   end
 
   @impl NimblePool
-  def handle_checkout(:checkout, _, %{mint: nil} = conn, pool_state) do
+  def handle_checkout(:checkout, _, %{mint: nil} = conn, {name, shp, _opts} = pool_state) do
     idle_time = System.monotonic_time() - conn.last_checkin
+    PoolMetrics.add(name, shp, in_use_connections: 1, total_checkout_count: 1)
```
sneako, line 177: I think if we stored the boolean `enable_pool_metrics?` in the pool state and check that before the `add/3` calls, it could be slightly faster in highly concurrent environments than the current method of checking if the pterm is present

sneako (commented):
Thank you so much for this contribution!

Running the test suite locally on main does require your patch for me as well so that is a welcome addition.

I have added some comments, please let me know what you think

oliveigah:
I agree with all suggestions! I'm currently on a business trip so probably won't be able to work on this until friday.

Regarding the averages and maxes, I think the potential margin of error is very low when we consider a large number of samples (which should be most cases).

The problem is if you take the measures without enough samples and therefore this margin of error may be more relevant.

One possible workaround would be only serve this measurements if there is enough samples, like 100. In this scenario even if the base metrics are slightly out of sync the real impact on the average/max measurements should not be relevant.

Anyway, I think is better to move on like you suggested with only available and in use connections and we can think more about this problem later.

Thanks for the feedback! (:

lib/finch/http1/pool.ex
```diff
@@ -16,36 +17,63 @@ defmodule Finch.HTTP1.Pool do
...
     )
   end
 
   @impl Finch.Pool
-  def request(pool, req, acc, fun, opts) do
+  def request(pool, req, acc, fun, name, opts) do
     pool_timeout = Keyword.get(opts, :pool_timeout, 5_000)
     receive_timeout = Keyword.get(opts, :receive_timeout, 15_000)
-
     metadata = %{request: req, pool: pool}
```
aselder, line 33: Since we're already passing the Finch name in, could we add it to the metadata so it's available in telemetry? It would be really useful in monitoring wait time per Finch instance?

sneako: Yes exactly @aselder I was thinking these changes enable #199 . I think that should be implemented in a new PR once this is complete.

oliveigah: I think this is a good idea indeed. But just to be clear, I've only added the finch_name on the Finch.Pool request callback because of averages and maxes metrics.

If we gonna move on only with available and in use connections this is not necessary.

Should I keep it here unused anyway in order to enable this change? What you think?

oliveigah: Considering the modifications on https://github.com/sneako/finch/pull/252 I think would be better to remove these changes from this PR in order to avoid conflicts.

oliveigah:
I'm working on the pool metrics implementation for HTTP2 and would like to know which metrics you think would be good to keep track of.

I'm thinking about in flight requests instead of available connections because it makes more sense on a multiplex scenario like HTTP2. There is another metric you think would be good @sneako?

oliveigah:
Regarding finch configurations with a pool count greater than one. Since the multiple pools are not exposed to the user in any way, I think makes sense to aggregate metrics from multiple pools under a single counter and generate metrics as there is only one pool.

An alternative would be keep separate metrics for each started pool using some index strategy and just show it as a unified metric when the user calls `Finch.get_pool_status` (maybe even have some option to show it separately). This would enable different forms of load balancing such as least_conn when there is more than one pool but the implementation may be harder.

What you think? I'm good with both. Just let me known the approach that better suits your vision for the library @sneako! (:

sneako:
Thanks @oliveigah !

I agree inflight requests make more sense for h2 :+1: 

I think individual pool metrics will be more useful than only exposing the aggregates. I definitely like the idea of enabling new load balancing strategies. Individual metrics will also be more useful when users are configuring their pool sizes & counts. Any aggregation I assume will take place somewhere else in the observability stack (ie Datadog or similar)

oliveigah:
I think the implementation is done. Gonna start to work on the docs now.

In summary:
- No more persistent_term call is needed when adding to the metrics, the atomic references are stored inside the pool state
- `Finch.get_pool_status/2` now returns a list of metrics struct. Each struct contains a `pool_index` attribute that represents the index of the pool.
- I've kept some other attributes on the pool state in order to facilitate future changes. This may causes some conflicts with https://github.com/sneako/finch/pull/252. Gonna wait for the merge to solve it.
- The only persistent term that still remains is the pool manager that keep all atomic refs related to a given finch instance and shp, this is used on the  `Finch.get_pool_status/2` call.
- Both h1 and h2 are ready

Let me know what you think @sneako. (:

lib/finch/http1/pool.ex
```diff
@@ -118,35 +137,57 @@ defmodule Finch.HTTP1.Pool do
...
+  @impl Finch.Pool
+  def get_pool_status(finch_name, shp) do
+    case Finch.PoolManager.get_pool_count(finch_name, shp) do
+      nil ->
+        {:error, :not_found}
+
+      count ->
+        result = Enum.map(1..count, &PoolMetrics.get_pool_status(finch_name, shp, &1))
+
+        if Enum.all?(result, &match?({:ok, _}, &1)),
```
sneako, line 149: Wouldn't partial results still be useful? For example, if a connection process were to restart, would that cause this function to return `{:error, :not_found}`? If so that error message could be misleading as pools do exist, but one or more happened to be unavailable at that moment. Plus, I think I would still like to get metrics for the rest of the pools in that case rather than none. Interested to hear your reasoning though.

oliveigah: I think you are right! My initial thought was that we probably never get into this state where some counters are not available. In fact with the current implementation the only way to this happen is on pool startup.

AFAICT is impossible to this situation happen otherwise because we are not cleaning (erasing de persistent_term) the state when the pool is terminated for any reason. So, even if we call `get_pool_status/2` between restarts the worst that could happen is serving stale metrics from the restarted pool until the new pool updates de persistent term on init.

Since I think would be desirable to leverage the terminate_pool callback that I've implemented on nimble_pool https://github.com/dashbitco/nimble_pool/pull/38 in order to do some clean up on pool termination, I think would be better to already go as you say and serve all the metrics we have even if some metrics are not available between restarts. Gonna fix this soon! Thanks for the feedback.

oliveigah:
Done @sneako! (:

Let me know if you see any other improvement to the proposed implementation.

sneako (approved):
Thank you @oliveigah!

PR #252, opened 2023-11-14
Add finch instance name to telemetry metadata

arthurnovak:
Recently, my team have adopted the practice of employing distinct finch instances for handling our requests. In this commit, we aim to include the pool-id/finch-instance-name in the telemetry metadata during request processing. This addition is particularly valuable for collecting statistics on a per-finch-instance basis.

This change aligns with both @aselder 's [request](https://github.com/sneako/finch/issues/199) and @oliveigah 's [PR](https://github.com/sneako/finch/pull/248), but it's a bit more concise. Hope this makes the review and merging process quicker.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -360,10 +360,9 @@ defmodule Finch do
           {:ok, acc} | {:error, Exception.t()}
         when acc: term()
   def stream(%Request{} = req, name, acc, fun, opts \\ []) when is_function(fun, 2) do
-    fun =
-      fn entry, acc ->
-        {:cont, fun.(entry, acc)}
-      end
+    fun = fn entry, acc ->
+      {:cont, fun.(entry, acc)}
+    end
 
     stream_while(req, name, acc, fun, opts)
   end
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -427,7 +427,7 @@ defmodule Finch do
 
   defp __stream__(%Request{} = req, name, acc, fun, opts) do
     {pool, pool_mod} = get_pool(req, name)
-    pool_mod.request(pool, req, acc, fun, opts)
+    pool_mod.request(pool, req, acc, fun, name, opts)
   end
 
   @doc """
@@ -560,7 +560,7 @@ defmodule Finch do
   @spec async_request(Request.t(), name(), request_opts()) :: request_ref()
   def async_request(%Request{} = req, name, opts \\ []) do
     {pool, pool_mod} = get_pool(req, name)
-    pool_mod.async_request(pool, req, opts)
+    pool_mod.async_request(pool, req, name, opts)
   end
 
   @doc """
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -40,12 +40,13 @@ defmodule Finch.HTTP1.Pool do
   end
 
   @impl Finch.Pool
-  def request(pool, req, acc, fun, opts) do
+  def request(pool, req, acc, fun, name, opts) do
     pool_timeout = Keyword.get(opts, :pool_timeout, 5_000)
     receive_timeout = Keyword.get(opts, :receive_timeout, 15_000)
     request_timeout = Keyword.get(opts, :request_timeout, :infinity)
 
-    metadata = %{request: req, pool: pool}
+    metadata = %{request: req, pool: pool, name: name}
+
     start_time = Telemetry.start(:queue, metadata)
 
     try do
@@ -55,9 +56,9 @@ defmodule Finch.HTTP1.Pool do
         fn from, {state, conn, idle_time} ->
           Telemetry.stop(:queue, start_time, metadata, %{idle_time: idle_time})
 
-          with {:ok, conn} <- Conn.connect(conn),
+          with {:ok, conn} <- Conn.connect(conn, name),
                {:ok, conn, acc} <-
-                 Conn.request(conn, req, acc, fun, receive_timeout, request_timeout, idle_time) do
+                 Conn.request(conn, req, acc, fun, name, receive_timeout, request_timeout, idle_time) do
             {{:ok, acc}, transfer_if_open(conn, state, from)}
           else
             {:error, conn, error} ->
@@ -90,7 +91,7 @@ defmodule Finch.HTTP1.Pool do
   end
 
   @impl Finch.Pool
-  def async_request(pool, req, opts) do
+  def async_request(pool, req, name, opts) do
     owner = self()
 
     pid =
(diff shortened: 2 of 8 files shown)


sneako:
Thanks for taking a look at this!

A few comments:
- Why only add it only to the `:queue` events and nothing else?
- Why only http1?
- Metadata is documented in the [Telemetry module](https://github.com/sneako/finch/blob/main/lib/finch/telemetry.ex) and should be updated accordingly.

arthurnovak:
Hey @sneako thank you very much for the review. I've updated `:queue` docs in telemetry module.

Regarding the other comments, I completely agree that it'd be great to report finch instance name everywhere. However, I'd prefer to make the rest of the updates in separate PRs so that the changes won't appear complex and can be quickly reviewed. Let me know if having only the queue update in this PR would work

samfrench (approved):

lib/finch/http1/pool.ex
```diff
@@ -26,12 +26,12 @@ defmodule Finch.HTTP1.Pool do
...
 
   @impl Finch.Pool
-  def request(pool, req, acc, fun, opts) do
+  def request(pool, req, acc, fun, name, opts) do
     pool_timeout = Keyword.get(opts, :pool_timeout, 5_000)
     receive_timeout = Keyword.get(opts, :receive_timeout, 15_000)
     request_timeout = Keyword.get(opts, :request_timeout, :infinity)
 
-    metadata = %{request: req, pool: pool}
+    metadata = %{request: req, pool: pool, name: name}
```
isavita, line 34: Here `metadata = %{request: req, pool: pool, name: pool_name}`

lib/finch/http1/pool.ex
```diff
@@ -26,12 +26,12 @@ defmodule Finch.HTTP1.Pool do
   end
 
   @impl Finch.Pool
-  def request(pool, req, acc, fun, opts) do
+  def request(pool, req, acc, fun, name, opts) do
     pool_timeout = Keyword.get(opts, :pool_timeout, 5_000)
     receive_timeout = Keyword.get(opts, :receive_timeout, 15_000)
     request_timeout = Keyword.get(opts, :request_timeout, :infinity)
```
isavita, line 32: here add `pool_name = Keyword.get(opts, :pool_name)`

lib/finch.ex
```diff
@@ -524,7 +524,7 @@ defmodule Finch do
   @spec async_request(Request.t(), name(), request_opts()) :: request_ref()
   def async_request(%Request{} = req, name, opts \\ []) do
     {pool, pool_mod} = get_pool(req, name)
-    pool_mod.async_request(pool, req, opts)
+    pool_mod.async_request(pool, req, name, opts)
```
isavita, line 527: The same as above `opts = Keyword.merge([pool_name: name], opts)`.

lib/finch.ex
```diff
@@ -391,7 +391,7 @@ defmodule Finch do
 
   defp __stream__(%Request{} = req, name, acc, fun, opts) do
     {pool, pool_mod} = get_pool(req, name)
-    pool_mod.request(pool, req, acc, fun, opts)
+    pool_mod.request(pool, req, acc, fun, name, opts)
```
isavita, line 394: What about having it `opts = Keyword.merge([pool_name: name], opts)`

isavita:
If we move the `pool_name` to `opts` we will not change the arity etc.

lib/finch.ex
arthurnovak: I like this one. It doesn't change arity of the funs and should make a step-by-step transition to use finch instance name in all events even smoother. @sneako thoughts?

arthurnovak:
hey @sneako can you give us any feedback please

lib/finch/pool.ex
```diff
@@ -4,11 +4,22 @@ defmodule Finch.Pool do
...
   @type request_ref :: {pool_mod :: module(), cancel_ref :: term()}
 
-  @callback request(pid(), Finch.Request.t(), acc, Finch.stream(acc), list()) ::
-              {:ok, acc} | {:error, term()}
+  @callback request(
+              pid(),
+              Finch.Request.t(),
+              acc,
+              Finch.stream(acc),
+              atom(),
```
sneako, line 12:
```suggestion
              Finch.name(),
```

lib/finch/pool.ex
```diff
@@ -4,11 +4,22 @@ defmodule Finch.Pool do
...
+              list()
+            ) :: {:ok, acc} | {:error, term()}
             when acc: term()
 
-  @callback async_request(pid(), Finch.Request.t(), list()) :: request_ref()
+  @callback async_request(
+              pid(),
+              Finch.Request.t(),
+              atom(),
```
sneako, line 20:
```suggestion
              Finch.name(),
```

sneako:
Thanks! I'd be ok with separate PRs for h1 & h2, but I'd rather not have separate PRs for each individual telemetry event.

The current implementation seems fine to me, I'm not too concerned about changing the arity of these internal/private behaviours and I don't feel it belongs in `opts` anyway.

isavita (approved):

arthurnovak:
hi @sneako have you had a chance to have a look at the change?

lib/finch/telemetry.ex
```diff
@@ -63,6 +63,7 @@ defmodule Finch.Telemetry do
 
   #### Metadata
 
+    * `:name` - The name of the Finch instance.
```
sneako, line 66: Thanks for this contribution! We should probably mention that name will only be populated for HTTP/1, but the more I think about it, I'd really like to get H2 done all at the same time.

arthurnovak: Hey, appreciate your review! I've included a note indicating that name is populated for HTTP/1 only in events utilised for both HTTP/1 and HTTP/2.

I believe beginning with the HTTP/1 change is a great step. It's a simple and direct modification that enhances the descriptiveness and usefulness of metrics. Plus, it can be easily expanded to HTTP/2 when the need arises.

sneako:
> hey @sneako, I had a bit of time around New Year's, so I went ahead and added the finch name to the telemetry meta for the H2 interface, as we discussed earlier. Could you please take a look? Oh, and by the way, happy new year!

Happy new year to you as well! Thank you for picking this up and completing the implementation. This looks good to me. Please run `mix format` and I will merge :)

arthurnovak:
> > hey @sneako, I had a bit of time around New Year's, so I went ahead and added the finch name to the telemetry meta for the H2 interface, as we discussed earlier. Could you please take a look? Oh, and by the way, happy new year!
> 
> Happy new year to you as well! Thank you for picking this up and completing the implementation. This looks good to me. Please run `mix format` and I will merge :)

done!

PR #259, opened 2024-02-07
fix timeout type

andyleclair:
Solves https://github.com/sneako/finch/issues/258


andyleclair:
Unrelated: I get these dialyzer failures when running dialyzer locally. I can take a whack at fixing it if you'd like? 
image

sneako (approved):

PR #261, opened 2024-02-08
fix types, appease dialyzer

andyleclair:
We need to not use `when is_reference` since the `ref` is technically an opaque type and we break that by using `is_reference` (or at least that's what i'm reading into dialyzer's output)


andyleclair:
I'd also add a caching step for the PLTs so we don't have to generate these every run, but I'm not very familiar with GH Actions (most of my experience is with CircleCI)

andyleclair:
@sneako it looks like the cache is working, I can see the dialyzer run took 4 seconds and didn't generate PLTs on the latest run

sneako:
Thank you!!

PR #267, opened 2024-02-24
Decrement in_use_connections when a connection is not reusable

### LINKED ISSUE #257: Reported `in_use_connections` grows beyond `pool_size` bounds
I'm using finch `0.17.0` working with HTTP1 connections and a pool configuration like this:

```elixir
pools: %{
  default: [
    size: 50,
    count: 1,
    pool_max_idle_time: :timer.seconds(120),
    conn_max_idle_time: :timer.seconds(120),
    start_pool_metrics?: true
  ]
}
```

`Finch.get_pool_status` is reporting an `in_use_connections` count that increases over time, eventually surpassing my pool size, which seems impossible.

Screenshot 2024-01-30 at 1 54 18 PM

_I'm guessing_ it's because I'm configuring a `conn_max_idle_time`, and any connections that exceed that limit when being checked out are closed, [but that closure isn't being recorded in the pool metrics](https://github.com/sneako/finch/blob/1a46841e1fccb99da4d9fec8e20dc4b998921d06/lib/finch/http1/pool.ex#L203-L215).

pcorey:
Fixes #257

### DIFF
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -212,9 +212,13 @@ defmodule Finch.HTTP1.Pool do
 
         Telemetry.event(:conn_max_idle_time_exceeded, %{idle_time: idle_time}, meta)
 
+        PoolMetrics.maybe_add(metric_ref, in_use_connections: -1)
+
         {:remove, :closed, pool_state}
 
       _ ->
+        PoolMetrics.maybe_add(metric_ref, in_use_connections: -1)
+
         {:remove, :closed, pool_state}
     end
   end


oliveigah:
> Thanks for looking at this @pcorey ! @oliveigah do you mind taking a quick look at this? I'm surprised this seems to fix the issue. To me it looks like this could cause the values to become negative, but maybe I'm not thinking this through enough :thinking: 

Sure! Gonna take a look at this today after work. When I have more info I will post it here.

sneako:
Thank you so much! No rush at all :pray:

oliveigah:
After gathering some context I agree with @sneako! It's not clear to me that this change solves the problem, in fact I did a test considering idle connection termination and this changes breaks it. The test is in this gist: https://gist.github.com/oliveigah/cbe7fa1ecb711a6006af5dd05c6803e7

The assert that breaks is the third call to `get_pool_status` :
```elixir
assert {:ok,
            [
              %PoolMetrics{
                pool_index: 1,
                pool_size: 10,
                available_connections: 2,
                in_use_connections: 8
              }
            ]} = Finch.get_pool_status(finch_name, shp)
```

The result with the proposed changes are: 
```elixir
{:ok, [%Finch.HTTP1.PoolMetrics{available_connections: 10, in_use_connections: 0, pool_index: 1, pool_size: 10}]}
```
This happens because when the connection is idle, every time the user makes a checkout 2 calls to `PoolMetrics.maybe_add/2` are made, one with -1 and other with 1 and in fact the metric is temporarily negative, and will be permanentlly negative after the checkin.

In summary, I think this change does not address the real problem we might have here.

About the root cause of the problem my thoughts are:

- The only way we can have an every increasing `in_use_connections` is by checkouts being completed without a checkin counterpart
- Reading through nimble pool's code https://github.com/dashbitco/nimble_pool/blob/deae8ffdb52ce5265a68299eea3e5074f1e8d5a7/lib/nimble_pool.ex#L432-L462 we may investigate if we are landing on any scenario that does not call `send(pid, {__MODULE__, :checkin, ref, client_state})` all of them are errors and might be noisy on your logs if you can share them with us @pcorey would be awesome

Gonna try to emulate some other scenarios to see if I can find any other possible root cause. @pcorey did this pool have a high throughput? I'm also suspecting about some weird interaction between metrics and pool termination, can you check if you have pool termination events in this pool?

pcorey:
Thanks for looking into this @oliveigah! To add to the confusion, my "fix", as you guessed, didn't actually work. I was conflating metrics from three different tenants. Two of those tenants are behaving normally; `available_connections` stays around 50 and `in_use_connections` stays around 0.

The third tenant is the one having issues. Here's a plot of it's `available_connections` (blue) vs `in_use_connections` (orange) in the 24 hours after a fresh deployment with mainline finch (`0.18.0`):

Screenshot 2024-03-18 at 6 07 36 AM

This tenant's Finch pool does experience higher load than the other two. The current request pattern is ~1 hour of 30-50 requests/minute followed by ~20 minutes of 1-5 requests/minute:

Screenshot 2024-03-18 at 6 16 02 AM

> all of them are errors and might be noisy on your logs

I'm not seeing any strange or unexpected errors in my logs. The only thing that may possibly be related is that all three of my tenants get connection closed errors from the upstream host (`%Mint.TransportError{reason: :closed}`).

> can you check if you have pool termination events in this pool?

I haven't had any pool termination events (`[:finch, :pool_max_idle_time_exceeded])` in the time range we're looking at.

oliveigah:
Thanks for the details @pcorey! 

The only problem I can see is some unexpected error on the `Mint.HTTP.request/5` call on `Finch.HTTP1.Conn` https://github.com/sneako/finch/blob/fb52686f9901ef7e1e04be03dd7b9ecb1549b686/lib/finch/http1/conn.ex#L115-L122

I've done a test that artificially raise when calling it and then I got the same behaviour you described! The test is here: https://gist.github.com/oliveigah/a7cf1c7931440ba6c647866fd3e99d74

The best way to solve this is to add a new callback to nimble pool when the request is canceled https://github.com/dashbitco/nimble_pool/blob/deae8ffdb52ce5265a68299eea3e5074f1e8d5a7/lib/nimble_pool.ex#L443-L444.

I've implemented a naive version of this new callback locally and it solves the problem. Gonna open an issue on nimble pool in order to gather some thoughts about this new callback.

Then we can use it here and see if it solves your problem! That's ok to you @sneako?

oliveigah:
PR was merged on nimble_pool. https://github.com/dashbitco/nimble_pool/pull/44

Since the maintainer asked for us try this fix before releasing a new version of nimble pool, I've implemented it on my fork https://github.com/oliveigah/finch using it from master instead of hex. 

All of my previously broken tests have been fixed with the changes, but if you could test it on your system and report the results would be awesome @pcorey! Let me know if I can help with anything.

pcorey:
@oliveigah I'm deploying with your fork now. I'll let you know how things look first thing Monday morning. Thanks for doing all of this!

pcorey:
@oliveigah Everything is looking great in my metrics. It looks like your fix worked! Should I close this PR and you can submit your own?
Screenshot 2024-03-25 at 9 36 35 AM

oliveigah:
Awesome @pcorey, thanks for the feedback!

Let's wait for a new nimble pool's relelase and then I can subimit the PR with the `handle_cancelled/2` implementation and the added tests.

oliveigah:
I've opened the new PR with the implementation of the new callback and added tests. I think we can cotinue this conversation over there. https://github.com/sneako/finch/pull/268

PR #268, opened 2024-03-25
feat: implement handle_cancelled/2 pool callback

oliveigah:


oliveigah:
@sneako, this version was tested on production by @pcorey and it solved the issue for him! (see the comments on https://github.com/sneako/finch/pull/267 for more details)

I've also added mimic (https://github.com/edgurgel/mimic) as a test dependency because it was the easiest way to emulate this bug on the unit tests. If you don't want to add it I can think about another way to achieve the same thing, just let me know.

Issue #265, opened 2024-02-20
Unable to send request with body larger than 64Kb


GRoguelon:
Hi,

I am not totally sure if this is a bug but I found a stranged behaviour in [Req](https://github.com/wojtekmach/req/issues/309) when I'm sending a body larger than 64Kb and it seems to be related to Finch `protocols` option.

It seems a change in the `protocols` provides different behaviours:

With `protocols: [:http1]`:
```elixir
iex(1)> {:ok, pid} = Finch.start_link(name: MyFinch, pools: %{ default: [ protocols: [:http1] ] })
{:ok, #PID<0.648.0>}
iex(2)> Finch.build(:post, "https://nghttp2.org/httpbin/post", [], :crypto.strong_rand_bytes(65538)) |> Finch.request(MyFinch)
{:ok,
 %Finch.Response{
   status: 200,
   body: "{\n  \"args\": {}, \n  \"data\": \"data:application/octet-stream;base64,TRUNCATED",
   headers: [
     {"date", "Tue, 20 Feb 2024 18:31:19 GMT"},
     {"content-type", "application/json"},
     {"content-length", "87690"},
     {"access-control-allow-origin", "*"},
     {"access-control-allow-credentials", "true"},
     {"x-backend-header-rtt", "0.156761"},
     {"strict-transport-security", "max-age=31536000"},
     {"connection", "close"},
     {"alt-svc", "h3=\":443\"; ma=3600, h3-29=\":443\"; ma=3600"},
     {"server", "nghttpx"},
     {"via", "1.1 nghttpx"},
     {"x-frame-options", "SAMEORIGIN"},
     {"x-xss-protection", "1; mode=block"},
     {"x-content-type-options", "nosniff"}
   ],
   trailers: []
 }}
```

With `protocols: [:http2]`:
```elixir
iex(1)> {:ok, pid} = Finch.start_link(name: MyFinch, pools: %{ default: [ protocols: [:http2] ] })
{:ok, #PID<0.648.0>}
iex(2)> Finch.build(:post, "https://nghttp2.org/httpbin/post", [], :crypto.strong_rand_bytes(65538)) |> Finch.request(MyFinch)
{:ok,
 %Finch.Response{
   status: 200,
   body: "{\n  \"args\": {}, \n  \"data\": \"data:application/octet-stream;base64,TRUNCATED",
   headers: [
     {"date", "Tue, 20 Feb 2024 18:27:43 GMT"},
     {"content-type", "application/json"},
[...]

PR #270, opened 2024-04-30
Use Mint.HTTP1 in HTTP1 conn

whatyouhide:
See https://github.com/sneako/finch/issues/265#issuecomment-1954896743.

### DIFF
--- a/lib/finch/http1/conn.ex
+++ b/lib/finch/http1/conn.ex
@@ -3,6 +3,7 @@ defmodule Finch.HTTP1.Conn do
 
   alias Finch.SSL
   alias Finch.Telemetry
+  alias Mint.HTTP1
 
   def new(scheme, host, port, opts, parent) do
     %{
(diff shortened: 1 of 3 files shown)


sneako:
Hey @whatyouhide !  Thanks for the PR. I guess we should also remove this part of the docs: https://github.com/sneako/finch/pull/250/files#diff-bfab6b899762596b9df3f3d7e719f23d973285f555d89dd956cfd59e652a7a25R60-R61

@wojtekmach want to make sure you see this, did you have any ideas for a workaround in Req?

wojtekmach:
Wait are we basically reverting https://github.com/sneako/finch/pull/250 or I'm missing something? It's fine if that is the plan i.e. we can't make it work but want to make sure.

sneako:
Yeah that is what would happen if we merge this PR... I do think Req should be able to support arbitrary protocols for unknown hosts, so I'm not sure if we should proceed this way. 

I guess the only alternative we have thought of so far is to "port HTTP/2 flow control in HTTP/1 connections" as @wojtekmach mentioned in the linked comment?

whatyouhide:
Okay, let's close this and figure out a good solution in #265 first.

Issue #272, opened 2024-05-14
Connection closed when request is attempted


salex89:
This might be a re-hash of these and other similar issues:
https://github.com/sneako/finch/issues/62
https://github.com/elixir-tesla/tesla/issues/467

However, I think something strange is happening, along with the other folks who commented on other issues, some very recent and mimic my experience. The issue is that certain requests fail with a `{:error, "socket closed"}` error, without actually being executed or at all and not due to any real connectivity issue. As a sidenote, the `{:error, "socket closed"]` result just the `{:error, %Mint.TransportError{reason: :closed}` returned by Finch wrapped by the Tesla adapter. But that's beside the point, that just says that the adapter is an unlikely culprit. 

These issues were much more common before I set the `pool_max_idle_time` to `60_000` down from the default `:infinity`, like 10 times more common. Yet the error still happens. I setup telemetry to try and gather more data on the issue. After capturing few such cases, this is the pattern I usually see (telemetry events forwarded to logs):

```
16:16:59.565 [info] Finch telemetry: reused_connection.
16:16:59.566 [info] Finch telemetry: http request sent, idle_time: 115862688122, method: DELETE, path: <redacted>
16:16:59.566 [info] Finch telemetry: receive: start, idle_time: 115862688122, method: DELETE path: <redacted>
16:17:00.749 [info] Finch telemetry: pool idle timeout, host: <redacted>, :pool_max_idle_time_exceeded.
16:17:00.750 [info] Finch telemetry: receive: stop, idle_time: 115862688122, method: DELETE path: <redacted>
16:17:00.750 <socket closed error here in my code>
...
<exponential backoff>
<retry succeeds with fresh connection with a sane idle_time>
```
Please notice two things, the `idle_time` is always a very large number when this happens, this is the equivalent of 3.7 years. Successful requests have an expected idle_time, less than 60 seconds. I also noticed (I believe always) that there is a `Finch telemetry: pool idle timeout` event when this error happens like immediately, a millisecond, before the connection closes. I didn't have telemetry from back when the `pool_max_idle_time` was at it's default (`:infinity`) but the the number of errors such was greater. It's a bit hard for me to revert it back to `:infinity` and test now easily because this code is deployed to an environment where it's being used and we need some traffic to make it happen, but if it would help debugging, I could.

We see this a few times each day towards different hosts, maybe 2-6 events daily, but on a mildly utilised system. We expect more traffic soon.

[...]

PR #273, opened 2024-05-22
Remove a connection if it receives any messages while in the pool

keathley:
Inspired by #272 I started looking into issues where idle connections are closed but not removed from the pool. Deep down I still feel like we need a packet capture to really suss out what's going on. Even if the pool hadn't processed a message telling us the tcp socket was closed, we should get an `einval` error when we attempted to put the connection into passive mode (at least in my testing on macOS).

That said, it occurred to me that a connection sitting in the pool should not be receiving data (since its finished the requests it needed to make and is back in the pool). Thus we can treat any message intended for that connection as an error and remove it from the pool.

Let me know what you think @sneako or if I missed something here.

### DIFF
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -241,7 +241,7 @@ defmodule Finch.HTTP1.Pool do
   @impl NimblePool
   def handle_info(message, conn) do
     case Conn.discard(conn, message) do
-      {:ok, conn} -> {:ok, conn}
+      {:ok, _conn} -> {:remove, :closed}
       :unknown -> {:ok, conn}
       {:error, _error} -> {:remove, :closed}
     end


sneako:
Hey @keathley ! Thank you so much for taking a lot at this. That definitely makes a lot of sense to me and seems pretty obvious in retrospect. 

I can deploy this PR to a production environment that makes extremely heavy use of Finch H1 pools and test out this hypothesis.

I'll report back with results ASAP!

sneako:
I deployed this to a single node in a cluster where I see occasional :closed errors.

The good news is that I am not noticing any negative side effects of this change.

On the other hand, there doesn't seem to be any difference in the rate of :closed errors. Below are two charts, one showing the :closed error count for the entire cluster, and the other showing the error count just on the test node.

The vertical gray bar just after 13:00 marks when I deployed this change to the test node.

image

sneako:
Definitely interested to see if this change helps anyone else who is seeing a higher rate of :closed errors. If anyone experiencing this issue could test out this PR and report back that would be greatly appreciated. As stated above, I can attest that no negative side effects were observed when I deployed this change to prod, so it should be a completely safe test to run for anyone interested.

elijahkim:
Hi! Coming from https://github.com/wojtekmach/req/issues/312#issuecomment-2145270023. I tried running this branch in a preview environment. I made an API request, waited a bit, and made another one which resulted in this error. I'm going to obfuscate some stuff but this is what I did.

```
Interactive Elixir (1.16.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> API.print("ID")
{:ok, <<37, 80,  ...>>}
iex(2)> NaiveDateTime.utc_now()
~N[2024-06-04 22:52:06.582967]
iex(3)> API.print("ID")

22:59:12.668 [error] retry: got exception, will retry in 1000ms, 3 attempts left

22:59:12.668 [error] ** (Req.TransportError) socket closed
{:ok, <<37, 80,  ...>>}
iex(4)> NaiveDateTime.utc_now()
~N[2024-06-04 22:59:30.035642]
iex(5)> API.print("ID")
{:ok, <<37, 80, ...>>}
iex(6)> NaiveDateTime.utc_now()
~N[2024-06-04 23:03:43.808728]
iex(7)> API.print("ID")

23:10:13.518 [error] retry: got exception, will retry in 1000ms, 3 attempts left

23:10:13.518 [error] ** (Req.TransportError) socket closed
{:ok, <<37, 80,  ...>>}
iex(8)> NaiveDateTime.utc_now()
~N[2024-06-04 23:10:21.185392]
iex(9)> API.print("ID")
{:ok, <<37, 80,  ...>>}
iex(10)> Hiive.Persona.Inquiries.print("ID")
{:ok, <<37, 80,  ...>>}
```

subsequent calls without waiting don't fail.

PR #274, opened 2024-06-10
Remove castore dependency

wojtekmach:
If we are on OTP 24, we'd receive this error:

image

Same error when using OTP 25+ but the OS happens not to have certs:

image

I believe the error is good enough however I'm happy to improve it of course!


sneako (approved):
Thanks @wojtekmach !

wojtekmach:
@sneako actually let's hold off on this for a second, perhaps we should set Mint version requirement to `>= 1.6.1`?

PragTob:
Less dependencies is always great :eyes: 

Any way to help this move along? Looks like one failing test :eyes:

wojtekmach:
The failing line is this one:

https://github.com/sneako/finch/blob/047a4177a8061ba971fb8398bfd7718554091300/test/finch/http2/pool_test.exs#L239

I have determined it started failing after this Mint change, https://github.com/elixir-mint/mint/pull/434, which makes sense, Mint stopped sending frames under some circumstances and the test is expecting these. Perhaps it's as simple as updating the test?

cc @v0idpwn

mix.exs
```diff
@@ -33,8 +33,7 @@ defmodule Finch.MixProject do
 
   defp deps do
     [
-      {:mint, "~> 1.6"},
-      {:castore, "~> 0.1 or ~> 1.0"},
+      {:mint, "~> 1.6.2 or ~> 1.7"},
```
wojtekmach, line 36: Mint is using OS trust store since 1.6.1. I thought I'd require the latest available version at the moment, 1.6.2.

PR #275, opened 2024-06-16
fix logo size

ruslandoga:
👋 

This PR restores the logo size to the same 350px height as before [the dark/light mode change.](https://github.com/sneako/finch/commit/3c7fdb287523265a746a4b19039584cfa7c84f5c)


sneako:
Thanks!

PR #278, opened 2024-07-20
fix: type Finch.request_opt() was missing the :request_timeout option

Mayeu:
Hello 👋 

First, thank you for your work on Finch 🙏

We encountered the following dyalixir error in our code:

```
lib/app/rpc.ex:352:no_return
Function post/2 has no local return.
```

Our function was only calling `Finch.build(...) |> Finch.resquest(...)` so nothing that could have raised. In this end, this was triggered because the `:request_timeout` keyword was missing from the `Finch.request_opt()` type.


Mayeu:
I forced push to fix the linting error the CI reported.

Issue #236, opened 2023-07-20
Finch.stream, again


hissssst:
Given what was said in #235, it would be nice to have a function

```elixir
Finch.stream(req)
```
Which would
1. Return a `Stream.t()`
2. And this stream would get lazily read from socket upon demand
3. Each item of the steam is `{:status, pos_integer()} | {:header, name :: binary(), value :: binary()} | {:body, binary()}`

josevalim:
Due to connection pooling and the state in the connection, it is not trivial to support this type of streaming. If you want lazy streaming, then `Finch.stream/5` is the best option for now. I think this can be closed as I am almost sure this was discussed in the past. :)

sneako:
Thanks for chiming in @josevalim !

hissssst:
@josevalim Yes, we've discussed, but the previous time the issue was closed due to new approach for streaming. This approach is discussed in #235 and it is an antipattern. I still demand for a proper streaming API, since existing solution with callbacks requires me to re-copy every part of the body to form a stream using send/receive, which is just unnecessary.

So, even if you don't have a solution right now, or not planning to implement it in the future, it is definitely worth to have this issue open to track progress and ideas if someone will have a time to implement it. @sneako , please, reopen.

## My first suggestion

1. The main problem here is that `NimblePool.checkout!` accepts a fun, so I'd suggest implementing separate `manual_checkout` and `manual_checkin` functions which a developer can use 

2. With manual checkout/checkin functions in NimblePool, we'll need to implement something like
```
socket = NimblePool.manual_checkout(...)
do_stream(conn)
|> Stream.concat(Stream.resource(fn -> :ok end, fn _ -> {:halt, _} end, fn _ -> NimblePool.manual_checkin(conn) end))
```

## My second suggestion

```
owner = self()
holder =
  spawn_link fn -> NimblePool.checkout!(..., fn conn ->
    send(owner, conn)
    receive do
      :release -> :ok
    end
  end)

conn =
  receive do
    conn -> conn
  end
  
do_stream(conn)
|> Stream.concat(Stream.resource(fn -> :ok end, fn _ -> {:halt, _} end, fn _ -> send(holder, :release) end))
```

It is missing timeouts, refs and other stuff. My point is just to share an idea

zachallaun:
I've been thinking about this since seeing these issues raised by @hissssst a few days ago and here are some thoughts:

[...]

PR #281, opened 2024-08-03
HTTP1 stream

hissssst:
This is a request for comments about initial HTTP1 streaming implementation from ideas in #236


lib/finch/http1/pool.ex
```diff
@@ -38,6 +38,112 @@ defmodule Finch.HTTP1.Pool do
...
+    metadata = %{request: req, pool: pool, name: name}
+
+    start_time = Telemetry.start(:queue, metadata)
+    owner = self()
+    ref = make_ref()
+
+    holder =
+      spawn_link(fn ->
+        try do
+          NimblePool.checkout!(
```
josevalim, line 56: This either needs to happen when the stream starts or you need to check later on that the process that checked out is the one the one that is streaming. Otherwise someone will pass the stream around to another process and it won't behave as expected.

lib/finch/http1/pool.ex
```diff
@@ -280,16 +386,22 @@ defmodule Finch.HTTP1.Pool do
...
-        end
-      else
-        {:ok, conn}
+          case Conn.transfer(conn, pid) do
+            {:ok, conn} -> {:ok, conn}
+            {:error, _, _} -> :closed
+          end
+
+        _ ->
+          {:ok, conn}
```
josevalim, line 404: Probably best to not change the code style to keep unrelated changes out of the pull request.

lib/finch/http1/conn.ex
```diff
@@ -100,6 +100,90 @@ defmodule Finch.HTTP1.Conn do
...
+                      if error do
+                        metadata = Map.merge(metadata, Map.put(resp_metadata, :error, error))
+                        Telemetry.stop(:recv, start_time, metadata, extra_measurements)
+                        raise error
+                      else
+                        metadata = Map.merge(metadata, resp_metadata)
+                        Telemetry.stop(:recv, start_time, metadata, extra_measurements)
+                      end
+
+                      send(handler, {ref, :stop, state})
```
josevalim, line 154: I don't see the connection being transferred back to the pool. Isn't that potentially an issue?

lib/finch/http1/pool.ex
```diff
@@ -38,6 +38,112 @@ defmodule Finch.HTTP1.Pool do
...
+              for connections. Consider adjusting the pool size, count, timeout or reducing the \
+              rate of requests if it is possible that the downstream service is unable to keep up \
+              with the current rate.
+              """,
+              trace
+            )
+
+          _ ->
+            exit(data)
+        end
```
josevalim, line 133: There is some code duplication with other functions, those should be extracted out.

lib/finch/http1/pool.ex
hissssst: > Otherwise someone will pass the stream around to another process and it won't behave as expected

I've tested it and it worked as expected, I'll push the test in a sec. Take a look at it please, perhaps I've misunderstood it.

lib/finch/http1/conn.ex
josevalim: Also notice that the after callback can be invoked in different situations:

1. The stream failed
2. Someone halted the stream

You need to recognize both scenarios accordingly. In the second one, the server can still be writing to the socket, and you need to cancel and stop that accordingly.

test/finch_test.exs
```diff
@@ -625,6 +625,43 @@ defmodule FinchTest do
...
+               {:status, 200},
+               {:headers,
+                [
+                  {"cache-control", "max-age=0, private, must-revalidate"},
+                  {"content-length", "2"},
+                  {"date", _},
+                  {"server", "Cowboy"}
+                ]},
+               {:data, "OK"}
+             ] = Enum.to_list(stream)
```
josevalim, line 661: We need to test several different scenarios:

1. What happens if the stream is halted before the response finishes?
2. What happens if the stream raises?

And we need to make sure the connection and the pool is still "functional" after that.

lib/finch/http1/pool.ex
josevalim: To be clear, it is probably best to do this lazily, if possible. As it is more flexible.

lib/finch/http1/conn.ex
hissssst: Nice catch, will fix

lib/finch/http1/pool.ex
hissssst: I agree, but I think that DRYing should be done after everyone agress on implementation detail

lib/finch/http1/conn.ex
hissssst: > In the second one, the server can still be writing to the socket, and you need to cancel and stop that accordingly.

Yes, connection is closed right below the comment `# In case some exception occured, we close the connection`

hissssst: Halting the stream also hits this branch, afaik

hissssst: The only case when connection is not closed, is them `{:done, mint_ref}` is returned by Mint. But assume that after this message, server is never writing to the socket

josevalim:
Btw, there may actually be a simpler implementation here.

The main insight here is to have one process hold the connection and the other do the streaming.

You can create any stream by returning an anonymous function that expects two arguments: the accumulator and one anonymous function. Therefore, actual stream could work like this:

```elixir
def actual_stream(...) do
  fn fun, acc ->
    conn = proxy_process_checkout!(...)
    Conn.request(..., acc, fun)
  end
end
```

And that would allow us to reuse most of the infrastructure that is already in place. The differences are:

1. Today `acc` is the value while in streams it is in the format `{:cont, acc}`, `{:halt, acc}`, `{:suspend, acc}`. However, you can change the existing implementation to pass `{:cont, acc}` tuples instead and implement the other formats.

2. Today the request function does not implement an `after` callback. Some code to execute when the stream halts or the invoked function raises. You could add one optional argument. So overall, the stream would do:

```elixir
def actual_stream(...) do
  fn fun, acc ->
    conn = proxy_process_checkout!(...)
    Conn.request(..., acc, fun, fn -> ...after... end)
  end
end
```

I think this direction would likely require fewer changes to the codebase and make sure that all request and stream implementations use the same code paths.

lib/finch/http1/pool.ex
josevalim: I will be bouncing out soon, but I meant this:

    stream = Finch.actual_stream(fn -> ... end)
    spawn(fn -> Enum.to_list(stream) end)

Even if this works today, because the process doing the streaming is not the one linked to, you could run into situations where the connection is never checked backed in, such as this:

    stream = Finch.actual_stream(fn -> ... end)
    pid = spawn(fn -> Enum.each(stream, fn _ -> Process.sleep(:infinity) end)
    Process.sleep(1000)
    Process.exit(pid, :shutdown)

hissssst: > you could run into situations where the connection is never checked backed in, such as this:

Unfortunately, I don't see any way to track this without explicit links. This limitation can be reflected in the documentation. But it is generally true for any possible resourse-oriented stream, like `File.stream!/3`, so

1. No possible implicit solution
2. True for any resouse-stream

Therefore, I wouldn't take any action except documentation for this one

josevalim: That's not true. You can guarantee this won't leak resources by using the `after` callback of `Stream.resource` and linking. You just need to make sure that linking is done in the first argument of `Stream.resource`, so it is done against the process actually consuming the streaming.

Another issue with this implementation is that it checks out too early. Checking out should be done as soon as the stream begins. Doing that would solve both problems.

hissssst: > so it is done against the process actually consuming the streaming.

Yeah, right, I haven't thought about it

hissssst: Did it

hissssst:
All done, I've also implemented failsafe check for stream which were lost and not started

test/finch_test.exs
```diff
@@ -625,6 +625,152 @@ defmodule FinchTest do
     end
   end
 
+  describe "actual_stream/3" do
+    test "Not supported for HTTP2", %{finch_name: finch_name} do
```
josevalim, line 629: Other test names start in lowercase, we should follow the convention.

test/finch_test.exs
```diff
@@ -625,6 +625,152 @@ defmodule FinchTest do
...
+
+      request = Finch.build("GET", endpoint(bypass))
+
+      assert {:ok, stream} =
+               Finch.actual_stream(request, finch_name,
+                 stop_notify: {self(), :stopped},
+                 fail_safe_timeout: 100
+               )
+
+      Process.sleep(200)
```
josevalim, line 764: Ideally we should avoid sleeping. Can we?

lib/finch/http1/pool.ex
```diff
@@ -38,6 +38,158 @@ defmodule Finch.HTTP1.Pool do
...
+    stop_notify = Keyword.get(opts, :stop_notify, nil)
+
+    metadata = %{request: req, pool: pool, name: name}
+
+    start_time = Telemetry.start(:queue, metadata)
+    owner = self()
+    ref = make_ref()
+
+    holder =
+      spawn_link(fn ->
```
josevalim, line 56: This is still checking out a connection and holding to it, before the streaming starts. You need to move this inside `fn {:cont, acc}, function ->`.

lib/finch/http1/pool.ex
```diff
@@ -38,6 +38,158 @@ defmodule Finch.HTTP1.Pool do
...
+                true ->
+                  Process.link(holder)
+
+                false ->
+                  raise "Process holding the connection in pool died" <>
+                          " before the stream enumeration started." <>
+                          " Most likely fail_safe_timeout occured"
+              end
+
+              try do
```
josevalim, line 126: The current code is not dealing with stream suspensions. Try implementing `Stream.zip(stream, stream)`. It needs to spawn two separate connections (instead of using the same one) and be able to yield out of this loop and come back in.

lib/finch/http1/pool.ex
josevalim: I have reviewed it and I like the new implementation a lot, it is much simpler. Great job! The only think remaining is making the checkout late and dealing with suspensions (which is used by `zip`). You have to be careful because a stream may not emit any item, be then suspended, and halted.

There are two ways you can do this: one is by moving the `after` callback, aka `send(holder, {ref, :stop, conn})`, to HTTP.conn. The flow of the code would be something like:


```
def stream(...) do
  # ...
  fn tagged_acc, fun ->
    conn = NimblePool.checkout!(...)

    HTTP1.request(tagged, fun, fn -> send(holder, {ref, :stop, conn}) end)
  end
end
```

Moving all error flow to HTTP.conn is simpler, because suspending would then just be something like this:

```
def request(..., {:suspend, acc}, function, after_fun) do
  {:suspended, acc, &request(..., &1, function, after_fun)}
end
```

In other words, streams are easier to implement if they are fully tail recursive, and not relying on try/catch and similar. Instead wrap each invocation of `fun` in try/catch.

lib/finch/http1/pool.ex
hissssst: > This is still checking out a connection and holding to it, before the streaming starts

Yeah, this is a design decision I've made on purpose.

If we perform checkout during stream start, we may end up in a situation, where stream was successfully created, but we can't iterate on it, because pool is busy and there's no free connection. And in this case, developer might be unable to retry the stream creation, because stream is data and it might've been already sent to another process and the request information can already be lost

On the other hand, current solution may lead to situations when connection was checked out, but no request was made since Stream was lost or enumeration has not started. I've implemented the fail-safe timeout for this situation specifically, to return connections to the pool

lib/finch/http1/pool.ex
hissssst: > The current code is not dealing with stream suspensions

Yeah, I don't understand what suspension means, that's why I decided to use `Stream.resourse` in the first place. I've read `Enum` and `Enumerable` documentation and haven't found any explanation of what halt and suspend mean and how to treat them

hissssst: Oh, I've found it, it's in type docs. I'll try to use it, thanks

lib/finch/http1/pool.ex
hissssst: > sidesteps both of these problems, because the connection and streaming happen immediately.

This is both true and not. Infinite recursion bug in `stream` callback will leave the connection checked-out forever, while `actual_stream` solves this problem.

> If the concern is retry, we could add the ability to retry inside the stream instead.

Yeah, this feels like a better solution, I agree. Check the connection on start and execute a callback or just retry with exponential backoff

josevalim: > Infinite recursion bug in stream callback will leave the connection checked-out forever, while actual_stream solves this problem.

How would you have an infinite bug in the stream callback? Are you saying in case our implementation has a bug? I am not sure those are valid arguments: a bug in `actual_stream` can cause connections to leak, eventually exhausting the pool and making the whole system unusable. I don't think we should use it as an argument against it either, we just need to make sure to address all cases.

hissssst: No, I mean something like
```elixir
{:ok, stream} = Finch.actual_stream(...)
Enum.each(stream, fn _ -> infinite_loop() end)
```

This can happen and connection will never be returned in pool. However, server will close the socket if it's unused for a long time, but I am not sure about this, since I am unaware if Mint sends empty ACK's to keep socket open

wojtekmach:
I wonder if it is possible to expose some of these internals publicly and create a stream on top of these. Say we have a LiveView and we use Finch.async_request today. My understanding is since it’s a firehose it could overflow the LiveView process to the point it’s unresponsive to handle_event and similar. If there is something like a blocking Finch.stream_next, returning something like :more/:done/etc, a LV could send a message to itself to stream next response chunk but otherwise be responsive to any other message. Or is there a selective receive for handle event somewhere to make sure it is handled first and this is mostly moot?

lib/finch/http1/pool.ex
hissssst: I've thought about it, and now we're back to the beginning where I've implemented it with `Stream.resourse`, but now I have to implement all of this suspend/halt logic plus change the request implementation

josevalim: Not really. The work is almost all done. Making it tail recursive is a matter of passing an after block to the existing code and adding one single clause to handle suspend. I still think it will be less code than the original PR. :)

hissssst:
@josevalim , I've implemented your suggestions, you can check the implementation. I didn't implement automatic retry though, but I think it's not a blocker. The only problem is that all failsafe timeout tests are passing until the last line and then dying with `exit shutdown` and I have no idea why

josevalim:
Thank you. I am swamped with ElixirConf, can you please remind me in 2 weeks if I did not comment before?

josevalim:
@hissssst your ping has been on my inbox since Oct 25th. I am aware of it, I just wanted to let you know it was not ignored. Sorry.

petejodo:
Prefacing this with that there's no expectations from my comment at all. From how this conversation ended 1.5 years ago, it seemed like this was close but may have just been forgotten and so I'm doing a ping. If @hissssst isn't around to resolve conflicts, I could take a stab at resolving them if there's still interest from the maintainer side.

hissssst:
I can resolve them, but I would love to have any reviews from maintainers before proceeding further. I think I have resolved all the existing comments, but some tests are flaky and it would be nice to get an extra pair of eyes there.

@sneako @josevalim , what do you think about this implementation?

hissssst:
I've resolved the conflicts.

There is one test which is flaky

And one test which doesn't work because `Finch.get_pool` returns `:not_found` for HTTP2 pool, which, I assume, is some bug inside Finch

PR #283, opened 2024-08-03
Document there is no backpressure on HTTP2

### LINKED ISSUE #282: Critical: HTTP2 stream will load whole response into memory
```elixir
Mix.install [:plug, :bandit, :finch]

defmodule P do
  @behaviour Plug
  import Plug.Conn

  def init(_ \\ []) do
    {:ok, []}
  end

  def call(conn, _) do
    conn = send_chunked(conn, 200)
    Enum.reduce(1..1_000_000_000, conn, fn _, conn ->
      {:ok, conn} = chunk(conn, String.duplicate("0", 1_000_000))
      conn
    end)
  end
end

Bandit.start_link(ip: {0, 0, 0, 0}, plug: P, port: 8000, http_1_options: [enabled: true])

Finch.start_link(name: TestFinch, pools: %{default: [size: 100, count: 1, protocols: [:http2]]})

r = Finch.build("GET", "http://localhost:8000/")
Finch.stream(r, TestFinch, [], fn _, _ ->
  IO.inspect Process.info(self(), :message_queue_len)
  Process.sleep(1_000)
end)
```

Run this and see in logs

```elixir
{:message_queue_len, 1}
{:message_queue_len, 15902}
{:message_queue_len, 33221}
{:message_queue_len, 50260}
{:message_queue_len, 67273}
{:message_queue_len, 84020}
```

Current implementation of streaming in HTTP2 is extremely dangerous since it may just OOM the caller, which is a bad thing to do. I'd mark http2 adapter as not ready for production use and raise in any client using it, until this gets fixed

josevalim:
Relates to #282.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -328,6 +328,15 @@ defmodule Finch do
 
   See also `stream_while/5`.
 
+  > ### HTTP2 streaming and back-pressure {: .warning}
+  >
+  > At the moment, streaming over HTTP2 connections do not provide
+  > any back-pressure mechanism: this means the response will be
+  > sent to the client as quickly as possible. Therefore, you must
+  > not use streaming over HTTP2 for non-terminating responses or
+  > when streaming large responses which you do not intend to keep
+  > in memory.
+
   ## Stream commands
 
     * `{:status, status}` - the http response status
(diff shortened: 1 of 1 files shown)


hissssst:
I think that it is also worth documenting for `async_stream` too. And overall approach must be pull model over push model, that means that `async_stream` needs to be reworked too

josevalim:
`async_request` already documents that it is a firehose. I think it is pretty much meant to be a fire and forget. If you want back-pressure, then I think pointing to one of the stream functions is fine.

sneako:
Thank you!

PR #287, opened 2024-08-05
Require Mint 1.6

wojtekmach:
This is in preparation for https://github.com/sneako/finch/pull/274.

As mentioned in https://github.com/sneako/finch/pull/274#issuecomment-2247267223, the test change is caused by upstream change in Mint.


test/finch/http2/pool_test.exs
```diff
@@ -235,9 +235,6 @@ defmodule Finch.HTTP2.PoolTest do
       ])
 
       assert_receive {:resp, {:ok, _}}
-
-      assert_recv_frames([rst_stream(stream_id: ^stream_id, error_code: :no_error)])
-
```
wojtekmach, line 240: @sneako tbh I'm not sure what this test is/was testing, it's not asserting on:

```elixir
assert_receive {:resp, {:error, %Finch.Error{reason: :request_timeout}}}
```

like other similar tests do.

wojtekmach: @zachallaun blast from the past! Could you help changing this test? Or if it outlived its usefulness, we let it go?

wojtekmach: Nevermind, you showed up in git blame but I think you were just moving code around, sorry about that!

sneako: I think its just testing the happy path? Seems fine to me to delete this assertion, let's do it

sneako:
Thanks again Wojtek!

test/finch/http2/pool_test.exs
wojtekmach: @sneako per:

```elixir
test "request timeout with timeout > 0 that fires after request is done"
```

it makes sense we don't assert on any returned errors because there aren't any. Perhaps this test was asserting the side effects (a frame with no_error from Mint) but  since there aren't any, perhaps it's not so useful anymore. Probably OK to to keep it though!

Issue #277, opened 2024-07-17
ArgumentError: unknown registry


tomtaylor:
We're seeing semi-regular stack traces in Sentry with the following:

```
ArgumentError: unknown registry: Obfuscated.FinchPool
  File "lib/registry.ex", line 1391, in Registry.info!/1
  File "lib/registry.ex", line 1007, in Registry.register/3
  File "lib/finch/http2/pool.ex", line 207, in Finch.HTTP2.Pool.init/1
  File "gen_statem.erl", line 984, in :gen_statem.init_it/6
  File "proc_lib.erl", line 241, in :proc_lib.init_p_do_apply/3
```

I think this is happening during our deployments when the app is being shutdown. Is it possible that the process tree in pool is being shutdown in the wrong order?

We're running finch `0.18.0` and mint `1.6.1` on Elixir 1.17.2. I think this has started happening since the upgrade to Elixir 1.17.

sneako:
I don't think that sounds likely. My best guess would be that your application is still trying to send requests after Finch has already been stopped. Does that sound possible?

sneako:
Closing due to inactivity, but if you have a more information that seems to contradict my theory feel free to comment.

elfenlaid:
Same for me, the same stack trace. This happens only on startups. I've never seen it on a local machine, but AppSignal reports this crashes regularly on production servers.

I assume it happens on startups because AppSignal attaches a hash of deployed changes to reported errors. The first instance of this error happens right after starting up a new version.

Happens after upgrading to Elixir/OTP 1.17.2/26.2.5.2 from 1.16.2/26.1.1. We were using Finch 0.18.0 for some time before upgrading the language version without errors.

Just to be certain nothing uses Finch before it's initialized, I start it up as the first application child. Still was getting the error. My configuration looks like this:

```elixir
    children = [
      # Start Finch
      {Finch,
       name: Mindme.Finch,
       pools: %{
         "https://api.stripe.com" => [protocols: [:http2]]
       }},
```

It could not even be related to Finch, we see similar errors start to pop up with other dependencies. So far, it happens to Mongo as well:

```
lib/gen_server.ex:1128 GenServer.call/3	
lib/mongo/auth.ex:34 Mongo.Auth.credentials/1	
lib/mongo/auth.ex:20 Mongo.Auth.run/2	
lib/mongo/mongo_db_connection.ex:131 Mongo.MongoDBConnection.connect/2	
[...]

PR #289, opened 2024-09-04
Tweak Finch supervisor children strartup order

elfenlaid:
Hello there, the changes aim to resolve the following issue: #277 

```elixir
ArgumentError: unknown registry: Obfuscated.FinchPool
  File "lib/registry.ex", line 1391, in Registry.info!/1
  File "lib/registry.ex", line 1007, in Registry.register/3
  File "lib/finch/http2/pool.ex", line 207, in Finch.HTTP2.Pool.init/1
  File "gen_statem.erl", line 984, in :gen_statem.init_it/6
  File "proc_lib.erl", line 241, in :proc_lib.init_p_do_apply/3
```

My comment here isn't correct – https://github.com/sneako/finch/issues/277#issuecomment-2265576904. The issue happens on shutdowns. 

---

The fix assumes that Supervisors shut down their children in reverse order. The problem with the current order is that `DynamicSupervisor` children use `Registry`. The registry goes down sooner than the dynamic supervisor does, causing the unknown registry exception.

I've tried to run the fix on the company servers, and I haven't encountered the unknown registry exception during deploys, at least so far. But maybe I've been lucky 😅 

Please let me know what do you think about the changes 🙏

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -184,8 +184,8 @@ defmodule Finch do
   @impl true
   def init(config) do
     children = [
-      {DynamicSupervisor, name: config.supervisor_name, strategy: :one_for_one},
       {Registry, [keys: :duplicate, name: config.registry_name, meta: [config: config]]},
+      {DynamicSupervisor, name: config.supervisor_name, strategy: :one_for_one},
       {PoolManager, config}
     ]
 


wojtekmach (approved):
So obvious in hindsight, good catch!

sneako:
Ah of course, thank you! I'll publish to hex asap

elfenlaid:
Thank you all for accepting the fix 🙇

Issue #291, opened 2024-09-26
NimblePool.checkout :idle_timeout When setting `pool_max_idle_time`


mathieurousseau:
Hello

We added a `pool_max_idle_time` to pool configuration.
We then started to have the following issue: 

```
Elixir.ErlangError Erlang error: "** (Oban.CrashError) ** (exit) exited in: NimblePool.checkout(#PID<0.26263.0>)\n    ** (EXIT) shutdown: :idle_timeout" 
    lib/finch/http1/pool.ex:96 Finch.HTTP1.Pool.request/6
    lib/finch.ex:493 anonymous fn/4 in Finch.request/3
    /app/deps/telemetry/src/telemetry.erl:324 :telemetry.span/3
    lib/req/finch.ex:239 Req.Finch.run_finch_request/3
    lib/req/finch.ex:71 Req.Finch.run/4
    lib/req/request.ex:1103 Req.Request.run_request/1
    lib/req/request.ex:1047 Req.Request.run/1
```

connection max idle time is 10_000
pool max idle time is 20_000

If we let it empty (:infinity). We have no issue. 

Thanks.
Mathieu

oliveigah:
Hey @mathieurousseau! I'm suspecting that this may be a issue of how idle termination is done on finch.

It stops the pool when the first connection becomes idle and not all of them. I'm not sure if this is the semantics we need here, and it's not clear how to achieve proper semantics with the current implementation of nimble_pool's idle ping feature. Gona take a look at it.

Is this bug consitently reproducible?

What I think may be happening is something like this:

- your pool have 2 connections idle for more than 20 seconds, so this pool should be terminated in the next verification cycle
-  before the pool is terminated some oban worker picks up the first connection to make some reequest
- now the idle verification happens and terminates the pool because the only remaining connection is idle for more than 20 seconds
- oban worker crashes because the finch pool is terminated with the reason idle timeout

TheOneric:
I recently got similar errors in my logs and switched finch to the patched version from #292 by @oliveigah for testing. *(Note: finch or nimble_pool were not upgraded before this started to occur, though another dependency also using NimblePool was upgraded; not sure if related)*

Unfortunately the patch did not seem to fix this as today this showed up in logs again:
```
2025-03-16 19:04:17.836 [error] #PID<0.26339.0> running MyApp.Web.Endpoint (connection #PID<0.26338.0>, stream id 1) terminated
Server: mydomain.example:80 (http)
Request: POST /inbox
[...]

PR #292, opened 2024-11-01
fix: include last checkout verification before idle termination

oliveigah:
closes https://github.com/sneako/finch/issues/291

Currently, idle verification stops the pool if the first connection exceeds the idle timeout. I don’t believe this is the correct behavior, as it may lead to bugs like those described in the issue.

The general approach here was to add a last checkout timestamp to the pool state and verify it in the handle_ping callback.

While I’m not keen on adding this to the pool state just for idle verification, I couldn’t think of a better solution. 

I’ll look into the possibility of adding a handle_ping callback specifically for the pool in nimble_pool, as the current handle_ping applies only to workers and doesn’t fully align with our use case. I think this would provide the semantics we need in Finch.

### DIFF
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -10,7 +10,8 @@ defmodule Finch.HTTP1.Pool do
       :shp,
       :pool_idx,
       :metric_ref,
-      :opts
+      :opts,
+      :activity_info
     ]
   end
 
@@ -189,7 +190,8 @@ defmodule Finch.HTTP1.Pool do
       shp: shp,
       pool_idx: pool_idx,
       metric_ref: metric_ref,
-      opts: opts
+      opts: opts,
+      activity_info: init_activity_info()
     }
 
     {:ok, state}
@@ -272,8 +274,12 @@ defmodule Finch.HTTP1.Pool do
   end
 
   @impl NimblePool
-  def handle_ping(_conn, %__MODULE__.State{} = pool_state) do
-    %__MODULE__.State{shp: {scheme, host, port}} = pool_state
+  def handle_ping(conn, %__MODULE__.State{} = pool_state) do
+    %__MODULE__.State{
+      shp: {scheme, host, port},
+      opts: opts,
+      activity_info: activity_info
+    } = pool_state
 
     max_idle_time = Map.get(opts, :pool_max_idle_time, :infinity)
     now = System.monotonic_time(:millisecond)
@@ -348,19 +354,28 @@ defmodule Finch.HTTP1.Pool do
   defp init_activity_info(),
     do: %{in_use_count: 0, last_checkout_ts: System.monotonic_time(:millisecond)}
 
-  defp update_activity_info(:checkout, pool_state) do
-    info = %{in_use_count: count} = elem(pool_state, 5)
+  defp update_activity_info(:checkout, %__MODULE__.State{} = pool_state) do
+    info = %{in_use_count: count} = pool_state.activity_info
 
-    put_elem(pool_state, 5, %{
-      info
-      | in_use_count: count + 1,
-        last_checkout_ts: System.monotonic_time(:millisecond)
-    })
+    %__MODULE__.State{
+      pool_state
+      | activity_info: %{
+          info
+          | in_use_count: count + 1,
+            last_checkout_ts: System.monotonic_time(:millisecond)
+        }
+    }
   end
 
   defp update_activity_info(:checkin, pool_state) do
-    info = %{in_use_count: count} = elem(pool_state, 5)
+    info = %{in_use_count: count} = pool_state.activity_info
 
-    put_elem(pool_state, 5, %{info | in_use_count: max(count - 1, 0)})
+    %__MODULE__.State{
+      pool_state
+      | activity_info: %{
+          info
+          | in_use_count: max(count - 1, 0)
+        }
+    }
   end
 end
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -218,7 +218,7 @@ defmodule Finch.HTTP1.Pool do
     with true <- Conn.reusable?(conn, idle_time),
          {:ok, conn} <- Conn.set_mode(conn, :passive) do
       PoolMetrics.maybe_add(metric_ref, in_use_connections: 1)
-      {:ok, {:reuse, conn, idle_time}, conn, update_last_checkout_ts(pool_state)}
+      {:ok, {:reuse, conn, idle_time}, conn, update_activity_info(:checkout, pool_state)}
     else
       false ->
         meta = %{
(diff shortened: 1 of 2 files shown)


oliveigah:
@mathieurousseau, I believe I’ve reproduced your issue in a unit test and have a solution in my fork.

If you could test it out in your code to see if it resolves the problem, that would be awesome!

sneako:
Thanks a lot @oliveigah !  I'll wait to see if @mathieurousseau says this resolved the issue.

oliveigah:
@sneako, based on @TheOneric’s report in #291, I've adjusted the approach slightly.

Now, I track two variables:

- Last checkout time
- Number of connections in use

This helps prevent issues when a connection remains checked out longer than pool_max_idle_time.

The main change was replacing the single last_checkout_time value in the pool state with a map that holds both pieces of data.

In the future, we might reuse the "in_use" data from metrics, but since metrics are optional, I needed to include this information in the pool state as well.

Let me know what you think, @sneako!

@TheOneric, let me know if the new version solves the issue for you!

TheOneric:
Thanks again for the patch! I can confirm no `:idle_timeout` errors showed up in my logs over the last week with this revised patch applied

lib/finch/http1/pool.ex
```diff
@@ -175,30 +175,34 @@ defmodule Finch.HTTP1.Pool do
...
     {:ok, _} = Registry.register(registry, shp, __MODULE__)
-    {:ok, {registry, shp, pool_idx, metric_ref, opts}}
+
+    {:ok, {registry, shp, pool_idx, metric_ref, opts, init_activity_info()}}
   end
 
   @impl NimblePool
-  def init_worker({_name, {scheme, host, port}, _pool_idx, _metric_ref, opts} = pool_state) do
+  def init_worker(
+        {_name, {scheme, host, port}, _pool_idx, _metric_ref, opts, _actt_info} =
```
sneako, line 184: at this point, i'm thinking its probably worth introducing a PoolState struct, the tuple is getting a bit complex. What do you think?

sneako:
Thanks again for your continued support @oliveigah ! 

I'm glad this change seems to resolve the issue. I have one small suggestion to consider, let me know what you think.

I'll make sure to stay on top of with this issue with you to finally get it resolved asap.

lib/finch/http1/pool.ex
oliveigah: I was thinking the same thing, but it will be a bigger change. 

I can do it on another PR or on this one. Let me know what you think and I can tackle it this weekend.

sneako: Ok yeah, another PR makes sense, thanks!

oliveigah: PR is open! https://github.com/sneako/finch/pull/308 as soon as it merges I will rebase this pr and fix the conflicts.

oliveigah:
Rebasing with the new pool state struct is done, @sneako! Let me know if any other changes are needed before merging this! Thanks!

TheOneric:
Today, running the version of this PR before the rebase unto the pool state refactor (i.e. af60cc38c6d42fd154d80d9c6a5872d3683a1953) I’ve encountered one more `:idle_shutdown` log; it’s the first such event since applying the patch:


```
2025-04-23 14:13:10.170 request_id=GDj4DPZIvKNRNJoAAeVS [debug] ---HTTP[a01322c1-dc28-4eac-9b67-3ed8d7dfe855]: Initiating request for: https://remote.domain.example/path

2025-04-23 14:13:10.206 request_id=GDj4DPZIvKNRNJoAAeVS [error] Internal server error: {{:shutdown, :idle_timeout}, {NimblePool, :checkout, [#PID<0.54506.0>]}}

2025-04-23 14:13:10.210 [error] #PID<0.54516.0> running MyApp.Web.Endpoint (connection #PID<0.54515.0>, stream id 1) terminated
Server: domain.example:80 (http)
Request: POST /inbox
** (exit) exited in: NimblePool.checkout(#PID<0.54506.0>)
    ** (EXIT) shutdown: :idle_timeout
```

The first debug message is created by a Tesla middleware just before the request is actually made, with Tesla using finch as the backend. The outgoing request was made in response to an incoming request and needed to finish processing the latter.

mathieurousseau:
> @mathieurousseau, I believe I’ve reproduced your issue in a unit test and have a solution in my fork.
> 
> If you could test it out in your code to see if it resolves the problem, that would be awesome!

Unfortunately I have moved away from the project for which we had this issue. 
@will-blosser could you take this forward?

oliveigah:
I'm coming back home today and may be able to take a look at this again on weekend. Sorry about the delay.

Gonna apply the suggested changes from @sneako and will do another investigation into the code to see any other possible way that idle termination of current checked out connections may happen as you experienced in your system @TheOneric 

Thanks for the feedback @mathieurousseau

@will-blosser if you could try the patches and give some more feedback it would be great.

oliveigah:
After reading through some parts of NimblePool's implementation, I believe idle termination errors like this might still occur due to how NimblePool monitors processes.

NimblePool calls this code from the client process (the process that checks out the connection):

https://github.com/dashbitco/nimble_pool/blob/64a3b081dc804bbe214ad739a7c5a5c2c298c086/lib/nimble_pool.ex#L445-L452

```elixir
  pid = GenServer.whereis(pool)

    unless pid do
      exit!(:noproc, :checkout, [pool])
    end

    ref = Process.monitor(pid)
    send_call(pid, ref, {:checkout, command, deadline(timeout)})
```
  
Since handle_checkout (implemented in Finch) is only executed inside the pool process, and the client process starts monitoring before the checkout happens, this may lead to the following scenario:

The pool is idle and sends a message for idle verification.

The client process starts monitoring the pool process.

The pool process consumes the idle verification message from its mailbox and, finding no connections in use, stops itself.

The client process raises an error due to the monitor.

I don't see a way to work around this with the current implementation. Since the activity information is stored inside the pool's state, we can only modify it via serialized calls to the process itself, which would hurt performance.

The only way I see to improve this would be moving to an implementation that can be efficiently parallelized — for example, using atomic counters, incrementing and decrementing the number of connections in use directly from the client process, and reading them from the pool process. This way, we could guarantee that if a process calls request/6, the pool won't be terminated by idle verification.

However, even with an atomics-based approach, it would still be possible for a race condition to occur, given the parallel nature of the problem. For instance, an atomic increment could happen after the idle verification check was already performed, still leading to idle termination errors.

In summary:

- We could move to a decentralized atomics implementation to reduce the occurrence of this bug even further, but it may still happen occasionally.

- I think implementing a full serialization (via process messages or locks) to guarantee 100% safety is not worth the performance and complexity cost, given how much of an edge case this scenario already is (requesting a connection from a pool that's about to be terminated).

Suggestions:

- @TheOneric I would suggest increasing the idle_time. I think 5 seconds is too low, and it's unlikely you need to be that aggressive in terminating idle pools. This would reduce the likelihood of hitting this edge case.

- An easy workaround would be a simple one-shot retry if an idle termination error occurs. This would effectively restart the just-terminated pool.

- To further reduce the chances of idle termination errors, a decentralized writes atomics-based implementation may be necessary.

- We already have an atomics approach in the current metrics implementation, although it isn't decentralized for writes yet — I could make that change.

- The metrics system could also be useful for issues like https://github.com/sneako/finch/issues/313 (e.g., autoscaling based on available connections) or for improving load balancing among pools.

- The main issue I see with making existing features such as idle termination dependent on metrics is that metrics are currently off by default, so we might have to change that behavior — enabling metrics automatically when idle termination is configured.

@sneako let me know what you think about all this so I can move forward with the implementation.

TheOneric:
> @TheOneric I would suggest increasing the idle_time. I think 5 seconds is too low, and it's unlikely you need to be that aggressive in terminating idle pools. This would reduce the likelihood of hitting this edge case.

Yes, this works for me; I was already planning to increase it but kept it low for now to keep conditions the same while testing the patch.  
Thank you for your continued work and investigation into this!

PR #295, opened 2024-11-22
Return acc when Finch.{stream,stream_while}/5 fails

jonatanklosko:
Currently `Finch.stream/5` and `Finch.stream_while/5` do not return accumulator in case of errors. If an error happens midway through streaming, there may be information in the accumulator that the user would like to access when handling the error.

This originally came up in https://github.com/wojtekmach/req/pull/435#discussion_r1853832157, where in the `{:status, 200}` handler we "open" collectable with `Collectable.into(collectable)` and put it in the accumulator. If the request fails, we want to call `collector.(:halt)`, however `collector` is only available in `acc`, and not returned on errors.

This changes the signatures as follows:

```diff
@spec stream(Request.t(), name(), acc, stream(acc), request_opts()) ::
-        {:ok, acc} | {:error, Exception.t()}
+        {:ok, acc} | {:error, Exception.t(), acc}
```

```diff
@spec stream_while(Request.t(), name(), acc, stream_while(acc), request_opts()) ::
-        {:ok, acc} | {:error, Exception.t()}
+        {:ok, acc} | {:error, Exception.t(), acc}
```

This is a breaking change, but I think it's justified, since it removes a constraint, thus making it more flexible for the user.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -369,7 +369,7 @@ defmodule Finch do
       File.close(file)
   """
   @spec stream(Request.t(), name(), acc, stream(acc), request_opts()) ::
-          {:ok, acc} | {:error, Exception.t()}
+          {:ok, acc} | {:error, Exception.t(), acc}
         when acc: term()
   def stream(%Request{} = req, name, acc, fun, opts \\ []) when is_function(fun, 2) do
     fun = fn entry, acc ->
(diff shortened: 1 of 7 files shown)


sneako:
Great idea, we definitely should have made it like this to start. We are still pre 1.0, so I think this breaking change, just to the stream functions is acceptable. 
Thanks @jonatanklosko!

jonatanklosko:
Fantastic, thanks :)

Issue #294, opened 2024-11-15
Discussion: support manual termination of the pools


sasa1977:
We are using Finch to communicate with dynamically started ephemeral servers. These servers have dynamic names, and so the set of URLs used grows indefinitely over time. 

Finch keeps the pool instances perpetually running for these servers, even after we don't need them. As a result, the amount of `NimblePool` processes also grows indefinitely. Setting the `:pool_max_idle_time` doesn't work. I am undecided whether this is a bug or a feature, but in any case, the idle option doesn't seem to work if the server is disconnected. 

Either way, I don't think idle timeout is an appropriate choice for this situation, partially because it only works on HTTP1, but perhaps more importantly, because it requires us to choose some magical timeout, instead of terminating the pool immediately after we don't need it anymore.

My first thought was to dynamically start a separate Finch instance per each dynamical server, and then stop it after it's not needed anymore. However, the instance name has to be an atom, so this means we would now be leaking atoms. This could be resolved if Finch supported arbitrary terms as names (which I think with some changes to the internals can be accomplished).

Another option, which seems to require much less work, is to support manual termination of the pool. I was able to accomplish this in the following way relying on the undocumented internals:

```elixir
{s, h, p, _, _} = Finch.Request.parse_url(url)

with {pid, _} = Finch.PoolManager.get_pool(MyFinch, {s, h, p}, auto_start?: false),
  do: DynamicSupervisor.terminate_child(MyFinch.PoolSupervisor, pid)
```

This could be wrapped in a function called e.g. `stop_pool`, which can be called by the clients when the pool is not needed anymore.

Thoughts?

sneako:
This sounds great to me! In my Finch usage we are hammering a static set of urls so this issue hasn't come up for us but I'm aware that using it to request arbitrary urls would result in the resource leak you mentioned.

I don't see any issue with introducing your proposal. PR is very welcome.

sasa1977:
Great, thanks! We'll try this idea on our project, and see if it works and solves our problems. If all goes well, we'll submit a PR.

sneako:
Closed in #299

PR #299, opened 2025-02-06
Support manual pool termination

sasa1977:
This PR introduces `Finch.stop_pool/2`, which can be used to manually stop the connection pool of a given SHP.

We have tried it out in production, and it works fine, and it helped us fix our process leak.

See #294 for details.

It's worth noting that this function might not be enough. The client needs to make sure that no request to the given SHP is taking place while the pool is being stopped. This can actually be tricky. A simple solution of channeling all requests through a single process means that you can only have one req at a time to the given SHP, which might be a problem.

In our case we've implemented a basic reference counted GC mechanism, which tracks all the processes currently making a request to the given SHP. When there are no requests taking place anymore, the pool is stopped.

I could be open to submit a PR for something like that, but the proper implementation will be much more complicated, because it has to support the differences between HTTP 1 and 2 implementations, especially for async requests.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -675,12 +675,17 @@ defmodule Finch do
   end
 
   def stop_pool(finch_name, shp) when is_tuple(shp) do
-    case PoolManager.get_pool(finch_name, shp, auto_start?: false) do
-      {pid, _pool_mod} ->
-        DynamicSupervisor.terminate_child(pool_supervisor_name(finch_name), pid)
-
-      :not_found ->
+    case PoolManager.all_pool_instances(finch_name, shp) do
+      [] ->
         {:error, :not_found}
+
+      children ->
+        Enum.each(
+          children,
+          fn {pid, _module} ->
+            DynamicSupervisor.terminate_child(pool_supervisor_name(finch_name), pid)
+          end
+        )
     end
   end
 end
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -656,4 +656,31 @@ defmodule Finch do
         {:error, :not_found}
     end
   end
+
+  @doc """
+  Stops the pool of processes associated with the given scheme, host, port (aka SHP).
+
+  This function can be invoked to manually stop the pool to the given SHP when you know it's not
+  going to be used anymore.
+
+  Note that this function is not safe with respect to concurrent requests. Invoking it while
+  another request to the same SHP is taking place might result in the failure of that request. It
+  is the responsibility of the client to ensure that no request to the same SHP is taking place
+  while this function is being invoked.
+  """
+  @spec stop_pool(name(), url :: String.t() | scheme_host_port()) :: :ok | {:error, :not_found}
+  def stop_pool(finch_name, url) when is_binary(url) do
+    {s, h, p, _, _} = Request.parse_url(url)
+    stop_pool(finch_name, {s, h, p})
+  end
+
+  def stop_pool(finch_name, shp) when is_tuple(shp) do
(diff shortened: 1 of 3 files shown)


sneako:
Thank you @sasa1977 !

It seems that this implementation will only shut down ONE pool, but Finch allows you to start many pools for a scheme/host/port.

What exactly happens if you shut down a pool with an ongoing request? I suppose we could consider sending a terminate message to each pool process instead of immediately telling the supervisor to terminate the child.

If a user if ok with sending all of their requests through a single process, then there is no reason for them to be using Finch in the first place, and they should probably just use Mint directly in that case.

I don't really like the idea of adding reference counting. It's probably minimal overhead, but even just a little bit of overhead is not worth it to support this feature IMO as it goes against the original design goals of the library.

sasa1977:
> It seems that this implementation will only shut down ONE pool, but Finch allows you to start many pools for a scheme/host/port.

Ah, this is a good catch! I could adapt the implementation to stop them all, assuming you're still ok with adding the support for this.

> What exactly happens if you shut down a pool with an ongoing request? I suppose we could consider sending a terminate message to each pool process instead of immediately telling the supervisor to terminate the child.

Not really sure, but I suppose the likely outcome is the failure of the request, or some kind of a process crash, since the pools are taken down while the request is running?

> If a user if ok with sending all of their requests through a single process, then there is no reason for them to be using Finch in the first place, and they should probably just use Mint directly in that case.

Finch provides additional value with a much simpler API. Another benefit is  that we can use req for more convenient goodies and features. Reimplementing all of this with Mint would require a lot of extra work, not to mention that the implementation would not be as mature, stable, or feature-rich as what Finch + req can give us.

Furthermore, in our case, we need this behaviour only for one some shps (essentially short-lived internal servers which are dynamically started in our cluster ala Flame). For other shps we don't stop the pool, and so we don't need to channel those requests through a single process.

Finally, it's worth noting that we're not channeling the actual requests through the process. Just some initial process registration is performed in this fashion (before the request is issued). The requests themselves are then issued concurrently, and then finally, deregistration is again channeled through the single process.

> I don't really like the idea of adding reference counting. It's probably minimal overhead, but even just a little bit of overhead is not worth it to support this feature IMO as it goes against the original design goals of the library.

That's fair enough, I just thought to mention it.

sneako:
Gotcha!

I'm definitely still on board with adding this feature as long as you update it to terminate all pools associated with a SHP.

I don't think it would be all too complicated to implement as send a message to all associated pools asking them to terminate, which would probably be more of a graceful shutdown, but I'm also fine with the currrent immediate termination approach as long as we mention in the docs that terminating a pool that is actively being used could cause some errors, if that sounds reasonable to you.

sneako:
Right of course, that's perfect.

sasa1977:
> I don't think it would be all too complicated to implement as send a message to all associated pools asking them to terminate

This could actually lead to subtle race conditions, because the pool might be stopped later when another request is started. So we really need a synchronous stop feature.

sneako:
> > I don't think it would be all too complicated to implement as send a message to all associated pools asking them to terminate
> 
> 
> 
> This could actually lead to subtle race conditions, because the pool might be stopped later when another request is started. So we really need a synchronous stop feature.

Good point! Ok let's roll with the sync route.

sasa1977:
> I'm definitely still on board with adding this feature as long as you update it to terminate all pools associated with a SHP.

Did it. Could you please take a look when you find the time?

sneako (approved):
Thanks @sasa1977 !

PR #307, opened 2025-03-27
adds :supported_groups to list of tls options

numso:
Here's a PR that fixes https://github.com/sneako/finch/issues/305 if we decide that's a desirable change.


sneako:
Thank you!

PR #308, opened 2025-03-29
chore: refactor http1 pool state

oliveigah:
As mentioned on https://github.com/sneako/finch/pull/292, in order to facilitate pool state handling would be good to create a new struct instead of a tuple.

### DIFF
--- a/lib/finch/http1/pool.ex
+++ b/lib/finch/http1/pool.ex
@@ -4,6 +4,7 @@ defmodule Finch.HTTP1.Pool do
   @behaviour Finch.Pool
 
   defmodule State do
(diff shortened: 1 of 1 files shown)


lib/finch/http1/pool.ex
```diff
@@ -3,6 +3,16 @@ defmodule Finch.HTTP1.Pool do
   @behaviour NimblePool
   @behaviour Finch.Pool
 
+  defmodule State do
```
sneako, line 6:
```suggestion
  defmodule State do
    @moduledoc false
```

sneako:
❤️

PR #314, opened 2025-04-14
Be more explicit about the `:default` pool

PragTob:
Until ~2 weeks ago all my co-workers and I were under the assumption that for not specified URLs a `:default` pool would be used, not that a new pool/pool group would be started for unknown origins/shp.

Hence, I aim to make this more explicit here as multiple people failed to grasp this (including myself :sweat-smile:).

Happy about feedback! And thanks for all your work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ children = [
   {Finch,
    name: MyConfiguredFinch,
    pools: %{
-     :default => [size: 10],
+     :default => [size: 10, count: 2],
      "https://hex.pm" => [size: 32, count: 8]
    }}
 ]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -51,15 +51,19 @@ children = [
...
      "https://hex.pm" => [size: 32, count: 8]
    }}
 ]
 ```
 
 Pools will be started for each configured `{scheme, host, port}` when Finch is started.
 For any unconfigured `{scheme, host, port}`, the pool will be started the first time
-it is requested. Note pools are not automatically terminated by default, if you need to
+it is requested using the `:default` configuration. This means given the pool
+configuration above each origin/`{scheme, host, port}` will launch 2 (`:count`) new pool
```
PragTob, line 63: This is more general and I can remove it again, but I think people may be familiar with the term of [origin](https://en.wikipedia.org/wiki/Same-origin_policy#Origin_determination_rules) and it may be useful in both the docs and maybe code base (aka calling the 3-tuple scheme, host, port origin).

sneako (approved):
❤️

PR #315, opened 2025-04-14
Upgrade `nimble_options` to document deprecations

PragTob:
The 1.1.1 release added notes for deprecations. This is helpful as the `:protocol` option was just hanging there, confusing me.

Drive by beautification with backticks.

Before patch/currently:

image

With patch:

image


sneako:
❤️

PR #329, opened 2025-09-17
Add support for querying default pool metrics

isavita:
Resolves sneako/finch#328

# Description 
- allow `Finch.get_pool_status/2` to accept `:default` and document the aggregated response shape.
- track default pool SHPs inside `Finch.PoolManager` so they can be queried and cleaned up when stopped.
- add `HTTP/1` tests to cover retrieving metrics for default pools and the not-found case.


isavita:
@sneako done here https://github.com/sneako/finch/pull/329/commits/836873c2509116680391adfe1c8e98a83a212390

PR #331, opened 2025-10-17
fix elixir 1.19 compilation warnings

sneako:


lib/finch/http1/pool.ex
```diff
@@ -381,7 +381,7 @@ defmodule Finch.HTTP1.Pool do
     }
   end
 
-  defp update_activity_info(:checkin, pool_state) do
+  defp update_activity_info(:checkin, %__MODULE__.State{} = pool_state) do
```
wojtekmach: I think might as well:

```elixir
defp update_activity_info(:checkin, %__MODULE__.State{} = pool_state) do
  update_in(pool_state.activity_info.in_use_count, &max(&1 - 1, 0))
end
```

PR #332, opened 2025-10-18
Add test case for issue #265 (ALPN + large body)

neilberkman:
Adds a test that reproduces the bug where POST requests with bodies >64KB fail when using protocols: [:http1, :http2].

The failing test is marked with @tag :skip. Two passing control tests confirm HTTP/2-only and HTTP/1-only configurations work correctly with large bodies.

Relates to #265


sneako:
Thank you!

PR #338, opened 2026-01-19
Encapsulate pool identifiers

NelsonVides:
I've started to do some work on https://github.com/sneako/finch/issues/296, and my idea is to allow for extra tags for a pool. The URL (and the `{schema, host, port}` that derives from it) is needed to identify a TCP connection, and therefore that we can't drop, so in order to distinguish two different TCP connections, my idea is to extend the identifier with "tags" (a tenant ID, a JWT, a customer name...). But before I can do that, the actual identifier I want to extend is scattered all over the repository and thought it'd be more incremental to separate encapsulating that from the rest of the work, as this is technically a refactor with no behavioural change, while the follow-up will have behaviour changes.

@sneako it's a bit noisy diff, but it's hopefully easy and a fine change. WDYT? I'm planning the follow up mentioned in the next few days :)

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -104,6 +104,8 @@ defmodule Finch do
 
   @type scheme_host_port() :: {scheme(), host :: String.t(), port :: :inet.port_number()}
 
+  @type pool_identifier() :: url :: String.t() | scheme_host_port() | Finch.Pool.t()
+
   @typedoc """
   Pool metrics returned by `get_pool_status/2` for a single pool.
   """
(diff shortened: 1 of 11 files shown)


sneako:
Thanks a lot @NelsonVides . I think this is a solid approach to prepare for the pool tag feature we need to enable separate pools to the same shp. 

I just merged https://github.com/sneako/finch/pull/337 which also touched shp, so could you please rebase and I'll get this merged in right away.

NelsonVides:
Rebased. Also took the chance to add specs to at least `PoolManager`, which is the most touched module here, I'm a big fan of specs and always regret they're not more ubiquitous in Elixir 🥲

When it comes to the flaky test, it seems to me like
```ex
    Task.async(fn -> delay_exec.(ref1, 10) end)
    Task.async(fn -> delay_exec.(ref2, 10) end)

    assert_receive {^ref1, :done}
    assert_receive {^ref2, :done}
```
Is expecting them in a certain order, which is what fails, the `:done` tags will come from different processes and therefore the order is not guaranteed, even if most of the times the first one to run will also be the first one to send. Isn't there a helper for receiving messages in any order... gonna check 🤔

NelsonVides:
Ah, no, wait, it's pinning the refs, it will check for exactly that. Nevermind, raising the timeout is my only idea for that one then :D

PR #342, opened 2026-01-24
Add pool tagging for connection pool isolation, and dynamic pool creation

### LINKED ISSUE #296: Support custom pool keys
I have an app that sends Apple Push Notifications on behalf of multiple different developer accounts.

Apple's APNs server requires any given connection to be used exclusively by the same authentication token.

https://developer.apple.com/documentation/usernotifications/establishing-a-token-based-connection-to-apns

> APNs doesn’t support authentication tokens from multiple developer accounts over a single connection.

Hence I can't use the URL schema as the pool identifier with Finch, I need a way to manually associate a connection pool with an authentication token. Essentially I need multiple separate pools which happen to connect to the same URL.

Is this something that could conceivably be supported in Finch?

NelsonVides:
Closes #296 and #345

This PR implements pool tagging and dynamic pool creation, enabling use cases like Req where
different connection options are needed for the same host at runtime. It enables using separate
pools to the same host for use cases where we want dedicated connections by tenant, JWT, or other
parameters.

### Public Finch.Pool Module

The `Finch.Pool` module is now public with a unified `new/2` constructor that takes parameters as
old, that is, either a URL or a Unix socket as it used to; and also takes a list of opts where
`:tag` is the only one supported for now:

- `Finch.Pool.new("https://api.example.com")` - creates a pool with default tag
- `Finch.Pool.new("https://api.example.com", tag: :api)` - creates a tagged pool
- `Finch.Pool.new({:http, {:local, "/tmp/socket"}}, tag: :api)` - Unix socket pools

### Static Pool Tagging

Pools can now be tagged to create separate connection pools for the same `{scheme, host, port}`
combination or Unix socket. This allows different configurations for 
different purposes (e.g., API vs web requests, tenants, JWT tokens).

**Configuration:**
```ex
  children = [
    {Finch,
     name: MyTaggedFinch,
     pools: %{
       Finch.Pool.new("https://api.example.com") => [size: 50, count: 4],
       Finch.Pool.new("https://api.example.com", tag: :web) => [size: 20, count: 2],
       :default => [size: 10, count: 1]
     }}
  ]
```

### Dynamic Pool Creation

Pools can now be started dynamically after Finch has been initialized using `Finch.add_pool/1`:

```ex
  Finch.add_pool(
    name: MyFinch,
    key: Finch.Pool.new("http://foo.com", tag: :hello),
    size: 30,
    count: 2
  )
```

### Usage

```ex
  request = Finch.build(:get, "https://api.example.com/users", [], nil, pool_tag: :api)
  Finch.request(request, MyTaggedFinch)
```

When a request specifies a `:pool_tag`, the tag must exist in the pool configuration. If it doesn't
exist, the request falls back to the `:default` configuration (not to a `:default` tagged pool for
the same SHP).

## Note

Also added protection against atom exhaustion by validating the finch instance existence before
creating supervisor registry atoms in `Pool.Manager.get_pool_supervisor/2`.

### Also included (from merged branches)

- **#351 – Unix sockets:** `http+unix://` and `https+unix://` URL scheme support; deprecation of `{scheme, {:local, path}}` in favour of URL strings.
- **#352 – Custom supervised pools:** `Finch.Pool` struct, `Finch.start_pool/3`, `Finch.find_pool/2`, `Finch.Pool.child_spec/1`, and user-managed pool lifecycle.

### DIFF
--- a/lib/finch/pool.ex
+++ b/lib/finch/pool.ex
@@ -156,8 +156,6 @@ defmodule Finch.Pool do
       raise ArgumentError, "expected :pool to be a Finch.Pool.t() struct, got: #{inspect(pool)}"
     end
 
-    # pool_opts = opts |> Keyword.drop([:finch, :pool])
-
     opts
     |> Keyword.drop([:finch, :pool])
     |> Finch.cast_pool_opts()


lib/finch/pool.ex
```diff
@@ -1,12 +1,30 @@
...
 
   @typedoc false
   @type request_ref :: {pool_mod :: module(), cancel_ref :: term()}
 
   @typedoc false
-  @opaque name :: {scheme(), host(), :inet.port_number()}
+  @opaque name :: {scheme(), host(), :inet.port_number(), pool_tag()}
 
   @doc false
   @callback request(
```
josevalim: What if we move those callbacks to the `Pool.Manager`? Otherwise we can accidentally leak types or callbacks. Keep only `from_name` and `to_name` private.

josevalim (commented):
I have added some comments. My only question is: do we want the pools to be managed inside Finch supervision tree or allow the user to manage it?

The current approach starts a pool inside supervision tree. But if we introduced `{Finch.Pool, ...}` in the PR, then the user would be responsible for hosting it, which would give them a bit more flexibility on how to architect the pool and when to shut it down. The implementation would be similar to this one, except we implement `Finch.Pool.child_spec` which will return:

```
def child_spec(opts) do
  Supervisor.child_spec({Finch.Pool.Supervisor, {name, {config.registry_name, pool, pool_config, track_default?}}},  opts)
end
```

lib/finch/pool/manager.ex
```diff
@@ -77,6 +83,26 @@ defmodule Finch.Pool.Manager do
     Registry.lookup(registry_name, :default)
   end
 
+  @spec start_pool_dynamic(atom(), Finch.Pool.t(), map()) :: :ok
+  def start_pool_dynamic(registry_name, pool, pool_config) do
```
josevalim: If we are going to keep this API, my suggestion is to build this tuple early:

```
{:via, Registry, {config.supervisor_registry_name, pool_name, data}}
```

And do a lookup in the registry before we sanitize pool config and build the other options, so in case the pool already exists, we don't attempt to call the dynamic supervisor and do unecessary work.

sneako:
Ah ok, so the new API was just meant for dynamic pools and users can continue to use the :pools option. That sounds great to me

README.md
```diff
@@ -63,6 +63,57 @@ it is requested using the `:default` configuration. This means given the pool
...
+You can configure tagged pools using `Finch.Pool.new/2`:
+
+```elixir
+children = [
+  {Finch,
+   name: MyTaggedFinch,
+   pools: %{
+     Finch.Pool.new("https://api.example.com") => [size: 50, count: 4],
+     Finch.Pool.new("https://api.example.com", tag: :web) => [size: 20, count: 2],
+     Finch.Pool.new({:http, {:local, "/tmp/api.sock"}}, tag: :api) => [size: 30, count: 2],
```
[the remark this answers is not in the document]
NelsonVides: Resolving this conversation to cleanup the state, we'll do this one in a separate PR

sneako:
I agree the scope of this PR has grown and planning to complete the discussed features in another PR makes sense.

In that case though we should probably remove `start_pool_dynamic/3` though right?

And then the new PR will be solely concerned with starting dynamic pools

josevalim:
> On this point, thinking, the idea sounds great. However, it seems like a scope creep in regards to this PR, should we prepare a separate issue for that?

In my mind, we wouldn't have both. We would only have the custom supervisor pools, so `add_pool` would be gone. In any case, I am fine with merging this as is, and we remove `add_pool` in the next PR. Let me do one final review now.

NelsonVides:
@josevalim @sneako, updated the PR description, now this branch contains all the recent work as merged. My only question is if we want to clean-up the git history before merging to main or it is ok as it is. Please give it one last review and let's move this forward :)

sneako:
I'm fine with the git history as is, great work!

Issue #279, opened 2024-07-24
More information on pool size vs. count


PragTob:
:wave: Hello there and thank you for `finch` and all the work around! :green_heart: 

I'd love some more information/documentation around the [pool configuration options](https://hexdocs.pm/finch/Finch.html#start_link/1-pool-configuration-options).

In particular my question is about pool size vs. pool count on HTTP1.

Like, what's the difference between having a pool size of 200 vs. a pool size of 50 and a pool count of 4 (which would also be 200 total connections). It's unclear to me how to optimize/balance these and am looking for some insight - and at best by updating the docs.

Looking at `NimblePool` I can gleam this from the downsides:

> The downside of NimblePool is that, because all resources are under a single process, any resource management operation will happen on this single process, which is more likely to become a bottleneck. This can be addressed, however, by starting one NimblePool per scheduler and by doing scheduler-based dispatches.

Meaning, the management of resources/checkouts may take too long if one pool was too big. Of course, hard to say what too big means.

I'd also be interested in knowing how a scenario with multiple pools is handled. Like, how is it decided which pool is used? Do we go round robin around the pools?

I think that information in the docs would be really appreciated, at least by me but also by others. Happy to try & write it myself - but for that I'd need to know about it :sweat_smile: 

## Background

I don't want to or need you to solve my problem, I know for that I'd go to elixirforum or the slack but I thought it might add some color to the issue for context for what I'm looking for. Feel free to ignore :)

I'm running essentially an "intelligent" proxy and so make a good chunk of HTTP requests. Roughly 180requests/second. Requests take between 50ms and ~2000ms (a small number times out after 15seconds, like 10). Roughly a quarter takes longer than 300ms.

I'm using finch via Req. First on this traffic (predictably) Finch failed with: 

> Finch was unable to provide a connection within the timeout due to excess queuing for connections. Consider adjusting the pool size, count, timeout or reducing the rate of requests if it is possible that the downstream service is unable to keep up with the current rate.

a couple of hundred times. Using ["Little's Law"](https://en.wikipedia.org/wiki/Little's_law) I estimated a queue size of ~72. Configured the pool size to 80 and count to 3 (so over provisioned a lot as that should give me 240).

Still got ~10 of the above errors (all at the same time) after that change which surprised me - as I thought I had sufficiently over provisioned the pools for the current traffic (I actually thought I configured it safely for ~2x the current traffic).

[...]

PR #361, opened 2026-03-20
Improve documentation around count/size and pool strategies

NelsonVides:
Hopefully this clarifies #279 (hello @PragTob! 👋🏽), as I wonder myself the same every single time I start a new Finch instance and tend to be confused again five minutes later 🙈

Note also that since the issue, we've added support for round-robin and hash-based strategies other than the random too.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -28,26 +28,45 @@ defmodule Finch do
       """,
       default: [:http1]
     ],
-    size: [
-      type: :pos_integer,
-      doc: """
-      Number of connections to maintain in each pool. Used only by HTTP1 pools \
-      since HTTP2 is able to multiplex requests through a single connection. In \
-      other words, for HTTP2, the size is always 1 and the `:count` should be \
-      configured in order to increase capacity.
-      """,
-      default: @default_pool_size
-    ],
     count: [
       type: :pos_integer,
       doc: """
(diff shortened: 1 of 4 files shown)


lib/finch.ex
```diff
@@ -28,26 +28,48 @@ defmodule Finch do
...
-      high pool count and few requests are made, these requests will be scattered \
-      across pools, reducing connection reuse. It is recommended to increase the pool \
-      count for HTTP1 only if you are experiencing high checkout times.
+      It is the maximum number of HTTP/1 connections per pool **worker**. Connections are opened
+      **lazily** up to this cap; the value is an upper bound, not a reservation. When every
+      connection in that worker is busy, further requests wait in the checkout queue until one is
+      returned or `:pool_timeout` (see `request/3`) is exceeded.
+
+      This applies only to HTTP/1 pools. For HTTP/2, a single connection multiplexes streams, so
+      per-worker size is effectively 1; use `:count` to run multiple HTTP/2 connections in parallel.
```
sneako:
```suggestion
      This applies only to HTTP/1 pools. For HTTP/2, this setting is ignored. A single connection
      multiplexes streams, so per-worker size is effectively 1; use `:count` to run multiple HTTP/2
      connections in parallel.
```

sneako (approved):
Great improvement!

Issue #313, opened 2025-04-14
[Feature Proposal] Start pools dynamically for a given origin


PragTob:
(I promise this is the last issue for today! 💚 Thank you etc. etc.!)

## Situation

Right now finch starts pools automatically when a new origin is encountered. The number of pools launched is static though (via `count`).

## Problem

It can be tough to figure out what pool count (and size) to configure, as mentioned in #279. Esp. when being new to finch (and/or Req for that matter) and not really understanding these concepts.

Esp. if you don't know in advance where your HTTP requests will go and how many there will be (somewhat niche use case, but not totally i.e. proxies or web hooks), it's hard to configure which leads you to over provision the `:default` with big numbers for pool size and count that may be unnecessary for 99% of origins.

## Solution

If we had an option that instead of raising when we couldn't get a connection we'd auto start a new pool to take care of that connection we could avoid these and allow pools to grow to their "natural"/needed size.

If #311 was fixed we could also allow pools to shrink again pool by pool (idle pools being shut down), i.e. the concern described there that if we shut down pools one by one we might end up with a too small pool count isn't a concern any more (as we'd then just start new pools again for the origin).

### Options:
* `initial_count` (or just `count`): This would be how many pools we'd initially start for a new origin
* `max_count`: I guess a concern is that we'd start way too many pools, so providing a `max_count` option is probably a good idea

## Implementation
In my head the implementation is somewhat simple? Just start a new one if we notice that the pool we selected is completely full. We could be more clever about this (#8) but I think it's good enough. Maybe the "realize pool is full/all connections are used" is more complicated though, I haven't looked yet.

As for selection strategy, right now we'd still end up over provisioning as pools are picked randomly so it's conceivable to end up with 10 pools each 5% full but since each one gets a new connection every now and then none would be shut down. That said, we can iterate on pool selection strategy or when to shut down a pool.

Also happy to try my hands at implementation, but again - I shouldn't be promising too much these days as my passion is usually bigger than my time 😂

sneako:
"Autoscaling" pools is definitely a feature I would be interested in. Since Finch started using lazy pools it *almost* behaves like this for my main use case, but obviously that doesn't work as well for the usage you describe (hitting uknown endpoints).

The code is already structured to easily allow for new pool selection strategies to be implemented, so that area is ready to be explored.

I agree that correctly implementing idle pool termination is probably important to finish before we are able to safely offer this feature.

[...]

Issue #312, opened 2025-04-14
Finch HTTP1 pool metrics aren't removed when the pool is


PragTob:
As usual, thanks a ton for all your work 💚 

## Problem

It seems to me that PoolMetrics aren't removed when the pool is terminated. That may make it seems like there are more pools than there actually are.

The following script demonstrates the problem and does the following:
* configures the default pools with a count of 3 and a timeout of 1 second
* does requests to 3 URLs, creating 3 x 3 = 9 pools
* looks at pool status and the actual registry in which Finch stores the pools, these match directly after making the requests
* then it sleeps for 5 seconds to kill off 3 pools (empty pools aren't shut down right now, see #311)
* then it prints the pool metrics and registry again, noting that the pool metrics still has 3 pools per origin while in the registry there are only 2 per origin 

```elixir

finch_name = Tobi.Tries
_finch = Finch.start_link(name: finch_name, pools: %{default: [count: 3, pool_max_idle_time: to_timeout(second: 1), start_pool_metrics?: true]})


urls = ["http://www.google.com", "https://www.screenversemedia.com/", "http://hex.pm"]

urls
|> Enum.map(fn url ->
  Task.async(fn -> Req.get!(url, finch: finch_name) end)
end)
|> Task.await_many()

urls
|> Enum.map(& (Finch.get_pool_status(finch_name, &1)))
|> IO.inspect()

urls
|> Enum.map(&Finch.Request.parse_url/1)
|> Enum.map(fn {s, h, p, _, _} -> {{s, h, p}, Registry.lookup(finch_name, {s, h, p}) |> length()} end)
|> IO.inspect()

IO.puts("\n\n\nSLEEEEEPPPYYYY TTTTIIIIMMMMEEEEEEEE\n\n\n")

# should kill the connections/pools
Process.sleep(to_timeout(second: 5))

urls
|> Enum.map(& (Finch.get_pool_status(finch_name, &1)))
|> IO.inspect()


urls
|> Enum.map(&Finch.Request.parse_url/1)
|> Enum.map(fn {s, h, p, _, _} -> {{s, h, p}, Registry.lookup(finch_name, {s, h, p}) |> length()} end)
|> IO.inspect()

```

The important output here is:

```
# ...

SLEEEEEPPPYYYY TTTTIIIIMMMMEEEEEEEE



[
  ok: [
    %Finch.HTTP1.PoolMetrics{
      pool_index: 1,
      pool_size: 50,
      available_connections: 50,
[...]

PR #362, opened 2026-03-21
Add runtime pool resize and replace persistent_term metrics with ETS

NelsonVides:
## Summary

- Add `Finch.get_pool_count/2` and `Finch.set_pool_count/3` for dynamically scaling pool workers at runtime without restarting the Finch instance
- Replace per-worker `persistent_term`/`atomics` metrics with a single ETS table per Finch instance, eliminating the global GC penalty of persistent_term writes when resizing pools
- Add `max_concurrent_streams` and `available_connections` to HTTP2 pool metrics
- Cleanup pool metrics when pools go down

## Motivation

Pool worker count was fixed at startup, requiring a full Finch restart to adjust capacity. During traffic spikes this means either over-provisioning upfront or accepting degraded performance.

The existing metrics implementation stored one `:persistent_term` entry and one `:atomics` array per worker. While fast for reads, this created many small entries that scale linearly with worker count and would make runtime resize problematic:
- removing a worker would remove its persistent_term entry, which would trigger a global GC on the entire node
- adding a worker would possibly trigger a full copy of the persistent-term table (that's how PT inserts work)

### **Runtime resize**

Uses the existing static Supervisor's `start_child/2`, `terminate_child/2`, and `delete_child/2` APIs. Worker count is derived from `Supervisor.count_children/1` rather than cached in the registry, ensuring accuracy after resize. The pool config is stored in the supervisor registry value so new workers can be built with the correct spec.

It kinda tackles #313 (hello @PragTob 👋🏽), the topic about autoscaling, though it rather just makes autoscaling possible instead of dealing with scaling algorithms and hysteresis and thresholds and all that dance here, leaving this library thin. Note that on #313, we've also already implemented custom strategies too.

### **Metrics rewrite**

Replaces persistent_term/atomics with a single named ETS table per Finch instance using `read_concurrency: true, write_concurrency: :auto`. Each worker stores one row keyed by `{pool_name, pool_idx}` with positional metric values. Protocol-specific modules (HTTP1/HTTP2 PoolMetrics) own the row layout and position mapping; `Finch.PoolMetrics` owns key construction and ETS operations. Workers store `{table, pool_name, pool_idx}` as their metrics ref, avoiding table name derivation on the hot path.

### **HTTP2 `max_concurrent_streams`**
Is read from Mint's server settings on connection and after every `HTTP2.stream/2` call, skipped entirely when metrics are disabled.

### **Metrics cleanup**
Happens on scale-down (per-worker `:ets.delete`) and `stop_pool` (bulk `:ets.match_delete`), preventing orphaned entries (this solves #312 – hello @PragTob 👋🏽)

### DIFF
--- a/lib/finch/pool/manager.ex
+++ b/lib/finch/pool/manager.ex
@@ -148,14 +148,8 @@ defmodule Finch.Pool.Manager do
           :not_found
 
         [{pid, {pool_mod, _pool_count, pool_config}}] ->
-          # Derive current worker count from supervisor children for accurate
-          # count after runtime resize via set_pool_count/3
-          try do
-            pool_count = Supervisor.count_children(pid).workers
-            {pid, pool_name, pool_mod, pool_count, pool_config}
-          catch
-            :exit, _ -> :not_found
-          end
+          pool_count = Registry.count_match(finch_name, pool_name, :_)
+          {pid, pool_name, pool_mod, pool_count, pool_config}
       end
     else
       :not_found
--- a/lib/finch/pool/manager.ex
+++ b/lib/finch/pool/manager.ex
@@ -148,7 +148,7 @@ defmodule Finch.Pool.Manager do
           :not_found
 
         [{pid, {pool_mod, _pool_count, pool_config}}] ->
-          pool_count = Registry.count_match(finch_name, pool_name, :_)
+          pool_count = Supervisor.count_children(pid).workers
           {pid, pool_name, pool_mod, pool_count, pool_config}
       end
     else
--- a/lib/finch/pool/manager.ex
+++ b/lib/finch/pool/manager.ex
@@ -136,7 +136,7 @@ defmodule Finch.Pool.Manager do
   end
 
   @spec get_pool_supervisor(Finch.name(), Finch.Pool.t()) ::
-          {pid(), pool_name(), module(), pos_integer()} | :not_found
+          {pid(), pool_name(), module(), pos_integer(), map()} | :not_found
   def get_pool_supervisor(finch_name, %Finch.Pool{} = pool) do
     pool_name = Finch.Pool.to_name(pool)
 
@@ -144,8 +144,18 @@ defmodule Finch.Pool.Manager do
     # This prevents atom creation when checking non-existent instances.
     if Process.whereis(finch_name) do
       case Registry.lookup(supervisor_registry_name(finch_name), pool_name) do
-        [] -> :not_found
-        [{pid, {pool_mod, pool_count}}] -> {pid, pool_name, pool_mod, pool_count}
+        [] ->
+          :not_found
+
+        [{pid, {pool_mod, _pool_count, pool_config}}] ->
+          # Derive current worker count from supervisor children for accurate
+          # count after runtime resize via set_pool_count/3
+          try do
+            pool_count = Supervisor.count_children(pid).workers
+            {pid, pool_name, pool_mod, pool_count, pool_config}
+          catch
+            :exit, _ -> :not_found
+          end
       end
     else
       :not_found
@@ -163,7 +173,7 @@ defmodule Finch.Pool.Manager do
     {:ok, config} = Registry.meta(finch_name, :config)
     pool_name = Finch.Pool.to_name(pool)
     pool_config = sanitize_pool_config(opts, pool)
-    data = {pool_config.mod, pool_config.count}
+    data = {pool_config.mod, pool_config.count, pool_config}
     name = {:via, Registry, {config.supervisor_registry_name, pool_name, data}}
 
     Supervisor.child_spec(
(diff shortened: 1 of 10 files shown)


lib/finch/pool/manager.ex
```diff
@@ -136,23 +136,33 @@ defmodule Finch.Pool.Manager do
...
-        [] -> :not_found
-        [{pid, {pool_mod, pool_count}}] -> {pid, pool_name, pool_mod, pool_count}
+        [] ->
+          :not_found
+
+        [{pid, {pool_mod, _pool_count, pool_config}}] ->
+          # Derive current worker count from supervisor children for accurate
+          # count after runtime resize via set_pool_count/3
+          try do
+            pool_count = Supervisor.count_children(pid).workers
```
sneako: What do you think about continuing to store the pool_count in the registry, just making sure to update it when we resize? I think typically get_pool_supervisor will be called much more often than resizing the pool

lib/finch/pool/supervisor.ex
```diff
@@ -2,33 +2,64 @@ defmodule Finch.Pool.Supervisor do
...
+        :ok
+
+      new_count < old_count ->
+        pool_name = Pool.to_name(pool)
+        table = PoolMetrics.table_name(registry_name)
+
+        for pool_idx <- old_count..(new_count + 1)//-1 do
+          Supervisor.terminate_child(sup_pid, pool_idx)
+          Supervisor.delete_child(sup_pid, pool_idx)
+          PoolMetrics.delete(table, pool_name, pool_idx)
```
sneako: Is it necessary to clean up the pool metrics both here and in the pool terminate callbacks?

lib/finch/pool/supervisor.ex
```diff
@@ -2,33 +2,64 @@ defmodule Finch.Pool.Supervisor do
...
+  end
 
-          spec ->
-            spec
+  def set_count(sup_pid, pool, registry_name, pool_config, old_count, new_count) do
+    cond do
+      new_count > old_count ->
+        for pool_idx <- (old_count + 1)..new_count do
+          spec = build_child_spec(pool, registry_name, pool_config, pool_idx)
+          Supervisor.start_child(sup_pid, spec)
```
sneako: do you think its worth having any logging or any kind of observability around the start and delete child calls failing?
if there happened to be concurrent calls to set_count, that could cause failures and it would probably be useful to signal that this is happening

lib/finch/pool/supervisor.ex
NelsonVides: Oh, right, this is duplicated 👌🏽

lib/finch/pool/manager.ex
NelsonVides: Just checked, and unfortunately, only the owning process can update that, which in this case it's a Supervisor, so we don't have any control over it other than the init callback, so this will be a super noisy change 🤷🏽‍♂️ 

https://hexdocs.pm/elixir/Registry.html#update_value/3

NelsonVides: Ah, but there's `Registry.count_match/3`, that'll do!

lib/finch/pool/supervisor.ex
NelsonVides: I cleaned the code a bit, and now just let the result of the Supervisor calls pop-up to the caller, so the caller might know what to do about them 🤔

lib/finch/pool/manager.ex
```diff
@@ -136,16 +136,20 @@ defmodule Finch.Pool.Manager do
...
     # This prevents atom creation when checking non-existent instances.
     if Process.whereis(finch_name) do
       case Registry.lookup(supervisor_registry_name(finch_name), pool_name) do
-        [] -> :not_found
-        [{pid, {pool_mod, pool_count}}] -> {pid, pool_name, pool_mod, pool_count}
+        [] ->
+          :not_found
+
+        [{pid, {pool_mod, _pool_count, pool_config}}] ->
+          pool_count = Registry.count_match(finch_name, pool_name, :_)
```
sneako: Not a bad idea, but this wouldnt include h2 connections that arent ready, right?

lib/finch/pool/supervisor.ex
```diff
@@ -2,33 +2,68 @@ defmodule Finch.Pool.Supervisor do
...
-            Map.put(spec, :significant, true)
+    Supervisor.init(specs, auto_shutdown: :all_significant, strategy: :one_for_one)
+  end
 
-          spec ->
-            spec
-        end
-      end)
+  @spec set_count(pid(), Pool.t(), atom(), map(), pos_integer(), pos_integer()) ::
+          :ok | {:error, {pos_integer(), term()}}
```
sneako: let's name these types too please

lib/finch/pool/supervisor.ex
```diff
@@ -2,33 +2,68 @@ defmodule Finch.Pool.Supervisor do
...
+            {:error, reason} -> {:halt, {:error, {pool_idx, reason}}}
+          end
+        end)
+
+      true ->
+        :ok
+    end
+  end
+
+  @spec build_child_spec(Pool.t(), atom(), map(), pos_integer()) :: Supervisor.child_spec()
```
sneako: and these types as well

lib/finch/pool/manager.ex
NelsonVides: Hmm... they don't register until ready so I think it would not include them... 😱 So what do you think, we take this tradeoff or we come back to the supervisor call? 😶

sneako: ok yeah i guess its fine since this only happens when getting pool status or shutting down a pool

sneako: @NelsonVides just to be clear, i meant i think the supervisor call is probably the right choice here. WDYT?

NelsonVides: Right, let's come back to this after the Easter break. Yes, the supervisor call is probably better here, gonna revert that 🤔

PR #364, opened 2026-04-02
Add graceful HTTP/2 connection draining via max_connection_age

### LINKED ISSUE #363: [Feature] Add HTTP/2 connection max-age draining for client-side load balancing
## Background

HTTP/2 multiplexes all requests over a single long-lived TCP connection. When the upstream service scales out — new nodes are added, instances are replaced, or a load balancer rotates its backend pool — existing Finch clients remain pinned to the original server endpoint for the lifetime of that connection. Traffic does not redistribute until the connection is torn down, either by the server or by a restart of the client application.

The conventional workaround is to downgrade to HTTP/1, which opens a new connection per request and therefore re-resolves DNS on each attempt. This sacrifices the performance and multiplexing benefits of HTTP/2.

This issue proposes a first-class solution: a configurable maximum connection age that causes pools to gracefully drain and reconnect. Each reconnect triggers a fresh name resolution, allowing the client to naturally rebalance traffic across all currently available upstream endpoints — without requiring a client restart or protocol downgrade.

---

## Proposed Solution

Add two new options to the pool configuration schema, and implement the draining lifecycle in `Finch.HTTP2.Pool`.

### New configuration options (`lib/finch.ex`)

| Option | Type | Default | Description |
|---|---|---|---|
| `max_connection_age` | `:timeout` | `:infinity` | Maximum time (ms) a connection stays open before draining begins. |
| `max_connection_age_jitter` | `:timeout` | `0` | Random additional duration added to the age limit to spread out reconnects and avoid thundering-herd spikes. |

### Pool draining lifecycle (`lib/finch/http2/pool.ex`)

The `Finch.HTTP2.Pool` `:gen_statem` is extended as follows:

1. **On connect** — start a state timer `{:timeout, :max_connection_age}` with duration `max_connection_age + rand(0..max_connection_age_jitter)`.
2. **On timer expiry** — unregister the pool from the Registry so no new requests are routed to it, set a `draining: true` flag, then either terminate immediately (if no in-flight requests) or transition to `:connected_read_only` (if requests are still in flight).
3. **On drain complete** — once the last in-flight request finishes while `draining: true`, the pool process terminates normally. The supervisor restarts it, triggering a fresh name resolution and a new connection.

### Transparent retry (`lib/finch.ex`)

If a request hits a pool that has just started draining, it receives `{:error, %Finch.Error{reason: :read_only}}`. The `__stream__/5` function is updated to catch this error and transparently retry the request; because the draining pool has already unregistered itself, the retry routes to an active pool (or causes a new one to be started).

---

## Acceptance Criteria

- [ ] `max_connection_age` and `max_connection_age_jitter` are accepted by `Finch.start_link/1` pool configuration.
- [ ] A pool whose age exceeds the configured limit unregisters itself from the Registry.
- [ ] In-flight requests on a draining connection are allowed to complete normally.
- [ ] New requests that land on a draining pool are automatically retried on a healthy pool, transparently to the caller.
- [ ] Over time, Finch distributes traffic across all endpoints returned by name resolution.

procmarchal:
Add graceful HTTP/2 connection draining via max_connection_age

Fixes #363

# Approach

Introduces two new `http2:` pool options — `max_connection_age` and `max_connection_age_jitter` — that let Finch proactively recycle HTTP/2 connections after a configured lifetime. This enables client-side load balancing against Kubernetes headless services, where long-lived HTTP/2 connections would otherwise stay pinned to a single pod.

When the age timer fires the pool unregisters itself from the Registry (so new requests are immediately routed elsewhere), finishes any in-flight requests in `:connected_read_only`, then terminates normally. The supervisor restarts it with a fresh DNS lookup. Requests that land on a draining pool via `request/3`, `stream/5`, or `stream_while/5` are transparently retried (up to 3 times); `async_request/3` callers receive the `:read_only` error and must retry manually (documented in the @doc).

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -105,7 +105,9 @@ defmodule Finch do
       doc: "HTTP/2-specific options. Only relevant when `protocols` includes `:http2`.",
       default: [
         wait_for_server_settings?: false,
-        ping_interval: :infinity
+        ping_interval: :infinity,
+        max_connection_age: :infinity,
+        max_connection_age_jitter: 0
       ],
       keys: [
         wait_for_server_settings?: [
@@ -127,6 +129,28 @@ defmodule Finch do
           the connection has been idle for this duration. When set to `:infinity` (default), \
           no PINGs are sent.
           """
+        ],
+        max_connection_age: [
+          type: :timeout,
+          doc: """
+          Maximum lifetime in milliseconds for an HTTP/2 connection before it is gracefully \
+          drained and replaced with a fresh one. When the timer expires the pool unregisters \
+          from the Registry (so new requests go to a fresh connection), finishes any in-flight \
+          requests, then terminates normally — the supervisor restarts it with a new DNS lookup. \
+          Useful for Kubernetes headless-service load balancing where DNS entries rotate. \
+          Defaults to `:infinity` (no age limit).
+          """,
+          default: :infinity
+        ],
+        max_connection_age_jitter: [
+          type: :non_neg_integer,
+          doc: """
+          Random jitter in milliseconds added to `:max_connection_age`. Prevents multiple \
+          pool shards from draining simultaneously (thundering-herd). The actual age used is \
+          `max_connection_age + :rand.uniform(max_connection_age_jitter)`. \
+          Defaults to `0` (no jitter).
+          """,
+          default: 0
         ]
       ]
     ]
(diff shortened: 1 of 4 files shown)


NelsonVides (commented):
All right! My first thought was that this is not a feature I'd use and therefore performance-wise didn't want to pay the price for it. But then I checked the code and well, I don't think there'd be any relevant performance penalty, the implementation is very clean! Then I thought well, wait, maybe actually _I do have_ a use case for this. Then I read the code again and thought: thank you!

Ok, first thing, could you rebase on top of the latest master and resolve conflicts? 🙏🏽

NelsonVides (approved):
LGTM! 🎉 

@sneako WDYT? 🙂

sneako:
Thanks @procmarchal and @NelsonVides !

PR #365, opened 2026-04-05
Validate keyword options in build/5 and request/3

sabiwara:
This is a suggestion in order to help catch typos or mistakes early.
For context, I spent some time debugging a timeout because I passed the `receive_timeout` to `build` instead of `request`.

I saw that `NimbleOptions` was used in other places, but I didn't want to go all the way in order to keep the change lightweight - happy to change it if it is the preferred way though.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -804,6 +804,7 @@ defmodule Finch do
 
   def request(%Request{} = req, name, opts) do
     validate_no_req_body_fun!(req, "Finch.request/3")
+    Keyword.validate!(opts, [:pool_timeout, :receive_timeout, :request_timeout, :pool_strategy])
 
     request_span req, name do
       acc = {nil, [], [], []}
--- a/lib/finch/request.ex
+++ b/lib/finch/request.ex
@@ -110,6 +110,7 @@ defmodule Finch.Request do
   @doc false
   @spec build(method(), url(), headers(), body(), build_opts()) :: t()
   def build(method, url, headers, body, opts) do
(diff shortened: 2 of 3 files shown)


NelsonVides:
Hey @sabiwara, thank you for your contribution!

Hmm... I love validation, but also, I love performance and this is in the hot path... I've also just tried with the type checker from Elixir 1.20 and it doesn't seem to me to catch it if I pass it some random atom, that would have been so ideal... 😭

Will come back to this in the next few days, maybe I'll benchmark something or so :)

sabiwara:
@NelsonVides being performance obsessed myself, I totally understand.
I also considered an alternative implementation that just raises but doesn't build any new list, and should have a much lower footprint.

```elixir
for {key, _} <- opts, key not in [...], do: raise ArgumentError, ...
```

`Keyword.validate!` has always felt a bit awkward to me for this reason, maybe we could add something in core for this (thinking aloud).

sabiwara:
@NelsonVides FYI I checked the code for `Keyword.validate/2` and found a potential optimization: https://github.com/elixir-lang/elixir/pull/15269. This should make it do less work in our case, since it can just return the list as is.

NelsonVides:
Wow, appreciate so much the effort to go the extra mile!

@sneako WDYT?

sneako:
Thank you @sabiwara ! I did purposefully avoid adding NimbleOptions here just to minimize work done on the hot path, but the Keyword.validate approach should be light enough, especially with your new optimization in core. Thanks a lot!

Issue #345, opened 2026-01-24
Support static and dynamic tagged pools


josevalim:
Hi everyone, Req needs to connect to multiple hosts with different options. Today Finch does support multiple hosts but they all must have the same configuration. Due to this limitation, Req may dynamically start new Finch trees when you pass different connection options. However, this comes with an issue that these trees are never garbage collected when idle. That's because Finch can garbage collect pools in a given tree, but Req is spawning multiple trees.

Therefore, we need a way for Finch to dynamically start pools (already supported) but with custom options (not currently supported). PR #338 from @NelsonVides did a good job cleaning up the codebase but there is more work ahead. Let's talk about the requirements first.

## Requirement A: static pool tagging

PR #342 introduces pool tagging:

```elixir
children = [
  {Finch,
   name: MyTaggedFinch,
   pools: %{
     {"https://api.example.com", :api} => [size: 50, count: 4],
     {"https://api.example.com", :web} => [size: 20, count: 2],
     {{:http, {:local, "/tmp/api.sock"}}, :api} => [size: 30, count: 2],
     {{:http, {:local, "/tmp/api.sock"}}, :web} => [size: 10, count: 1],
     :default => [size: 10, count: 1]
   }}
]
```

Unfortunately Req needs to start those dynamically. This means the configuration options are not known on boot, only at runtime.

## Requiring B: dynamic pool creation

The idea is to allow pools to be started dynamically and explicitly:

```elixir
Finch.Pool.start_link(
  key: Finch.Pool.from_url("http://foo.com", tag: :hello),
  size: 30,
  count: 2
)
```

The `:tag` is optional. Now, whenever you build the request, you can pass the pool tag:

```elixir
Finch.build(..., pool_tag: :hello)
```

The difference is, if a `pool_tag` is given, **it must match an existing entry in the pool** (which is different from #342).

One challenge of this solution is, what happens if two processes try to start a pool with the same tag? Currently, the `PoolManager` process acts a single entity which could be a bottleneck and may become a larger bottleneck with explicit pool creation. This is fixed in #344.

## Next steps

Here is my proposal to address this:

  * [x] Stop passing `{scheme, host, port}` around (done in #338)
  * [x] Provide an explicit boundary between pool definitions and pool names (done in #344)
  * [ ] Introduce the :tag configuration and make `Finch.Pool` public (see below)
  * [ ] Introduce `Finch.Pool.start_link`

## Introduce the :tag configuration and make `Finch.Pool` public

[...]

PR #366, opened 2026-04-12
Add Elixir Registry support for Finch instance naming

meox:
### Problem

Finch only accepts `atom()` for the `:name` option in `start_link/1`. This forces
users who dynamically create Finch pools from runtime configuration (e.g., config
files, databases, multi-tenant setups) to use `String.to_atom/1`, which leaks atoms
since the BEAM never garbage-collects them. This is a well-known footgun for any
system with unbounded dynamic names.

### Solution

Allow `{:via, Registry, {registry, key}}` tuples as the `:name` option, following
the standard Elixir convention for dynamic process naming without atom creation.

```elixir
# Before (atom leak from dynamic input):
name = String.to_atom("finch_#{tenant_id}")
Finch.start_link(name: name)
Finch.request(req, name)

# After (safe, no atom creation from user input):
name = {:via, Registry, {MyApp.FinchRegistry, tenant_id}}
Finch.start_link(name: name)
Finch.request(req, name)
```

### Design

Finch's internals (Elixir `Registry.lookup/2`, named ETS tables) require atom
identifiers — this is a hard constraint from the BEAM and Elixir's `Registry`
module (`@type registry :: atom`). The solution introduces a thin indirection
layer:

1. **`Finch.NameRegistry`** — a shared ETS table (`:finch_via_names`) that maps
   `{:via, Registry, {reg, key}}` tuples to auto-generated internal atoms
   (e.g., `:"Finch.Instance.42"`). These atoms are bounded by the number of
   active Finch instances, not by user input.

2. **`Finch.ViaCleaner`** — a tiny GenServer added as a child of the Finch
   supervision tree (only when using via tuples) that removes the ETS mapping
   entry on termination, preventing stale entries.

3. **`resolve_finch_name/1`** — called at the entry point of every public API
   function. For atoms it's a no-op passthrough; for via tuples it resolves to
   the internal atom via the ETS lookup.

### What changes

| File | Change |
|------|--------|
| `lib/finch/name_registry.ex` | **New** — ETS-backed via→atom mapping (register/resolve/unregister) |
| `lib/finch/via_cleaner.ex` | **New** — GenServer that cleans up ETS entry on shutdown |
| `lib/finch.ex` | Update `@type name()`, `start_link/1`, `init/1`, `finch_name!/1`; add `resolve_finch_name/1`; wire resolution into all public API functions |
| `lib/finch/pool/manager.ex` | Fix race in `get_pool_supervisor/2`: catch `:noproc` exit when supervisor dies between Registry lookup and `Supervisor.count_children/1` |
| `test/finch_test.exs` | New `"Registry-based naming"` describe block with tests |
| `test/finch/http1/pool_test.exs` | Fix flaky idle-timeout tests: increase `assert_receive` timeouts from 100-150ms to 500ms |

### What stays unchanged

- All internal modules (`Pool.Manager`, `Pool.Supervisor`, `HTTP1.Pool`,
  `HTTP2.Pool`, `PoolMetrics`) — they only ever receive atoms after resolution
- All existing atom-based naming — 100% backward compatible
- Zero performance impact on atom-named instances (`resolve_finch_name/1` for
  atoms is a single pattern match)

### Tests

- Start Finch with via tuple, make HTTP/1 request
- Multiple via-tuple instances coexisting
- `get_pool_status/2` with via tuple name
- Dynamic `start_pool/3` with via tuple name
- `stop_pool/2` with via tuple name
- `get_pool_count/2` and `set_pool_count/3` with via tuple name
- ETS cleanup: entry removed after Finch stops
- Supervisor discoverable via user's Registry
- Validation: raises on invalid name shape

### Flaky test fixes

- **`get_pool_supervisor/2` race condition** — `Supervisor.count_children/1` was
  called on a PID obtained from `Registry.lookup/2` without guarding against the
  supervisor dying between lookup and call. Fixed by catching `:noproc` exits and
  returning `:not_found`.

- **HTTP1 pool idle timeout tests** — `assert_receive` timeouts of 100-150ms were
  too tight under parallel test load. Increased to 500ms. The `refute_receive`
  timeouts (which test that something does NOT happen within a window) are left
  unchanged since they need to be tight to be meaningful.

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -134,8 +134,13 @@ defmodule Finch do
 
   @typedoc """
   The `:name` provided to Finch in `start_link/1`.
+
+  Can be an atom (traditional) or a `{:via, Registry, {registry, key}}` tuple
+  for dynamic naming without atom creation. When using a via tuple, Finch
+  auto-generates an internal atom for its subsystems; users never need to
+  call `String.to_atom/1`.
   """
-  @type name() :: atom()
+  @type name() :: atom() | {:via, Registry, {atom(), term()}}
 
   @type scheme() :: :http | :https
 
@@ -334,19 +339,31 @@ defmodule Finch do
   """
   def start_link(opts) do
     name = finch_name!(opts)
+
+    {internal_name, supervisor_name, via_name} =
+      case name do
+        atom when is_atom(atom) ->
+          {atom, concat_name(atom, "Supervisor"), nil}
+
+        {:via, Registry, {_reg, _key}} = via ->
+          internal = Finch.NameRegistry.register(via)
+          {internal, via, via}
+      end
+
     pools = Keyword.get(opts, :pools, []) |> pool_options!()
     {default_pool_config, pools} = Map.pop(pools, :default)
 
     config = %{
-      registry_name: name,
+      registry_name: internal_name,
       registry_listeners: Keyword.get(opts, :registry_listeners, []),
-      supervisor_name: Pool.Manager.supervisor_name(name),
-      supervisor_registry_name: Pool.Manager.supervisor_registry_name(name),
+      supervisor_name: Pool.Manager.supervisor_name(internal_name),
+      supervisor_registry_name: Pool.Manager.supervisor_registry_name(internal_name),
       default_pool_config: default_pool_config,
-      pools: pools
+      pools: pools,
+      via_name: via_name
     }
 
-    Supervisor.start_link(__MODULE__, config, name: concat_name(name, "Supervisor"))
+    Supervisor.start_link(__MODULE__, config, name: supervisor_name)
   end
 
   @doc """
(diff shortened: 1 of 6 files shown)


NelsonVides (commented):
Question, this is intended to help with the case when you create dynamic Finch instances at runtime, but, wouldn't a single instance with dynamic pools do the work?

We've recently introduced this feature, it's already in `main` and to be released to hex soon enough. This PR introduces quite some complexity so I'd like to understand what does this solve that the new recent features don't 🤔

meox:
the main goal is to start the pool at runtime without generating atoms

NelsonVides:
Yeah, but you can start pools dynamically at runtime without generating atoms already, see https://github.com/sneako/finch/issues/345 and all the linked PRs, doesn't that exactly solve the issue already or is there anything missing? 🤔

meox:
> Yeah, but you can start pools dynamically at runtime without generating atoms already, see #345 and all the linked PRs, doesn't that exactly solve the issue already or is there anything missing? 🤔

I missed that part so probably we have to adapt the Req lib to take advantage of that?

NelsonVides:
It hasn't made it to a hex package yet, I think we're just missing out a Mint release here https://github.com/sneako/finch/pull/341, merge that, and Finch should be ready. Jose and Wojtek are aware of these changes (Jose proposed them!) so Req is probably adapting to this too already, but I haven't followed Req's changes :)

Thanks a lot for your PR nevertheless!

PR #368, opened 2026-04-17
Pool metrics improvements

NelsonVides:
Somewhat a follow up on https://github.com/sneako/finch/pull/362, having the pid on the metrics could help me correlate load to processes. Also, reworked the implementation to use `ordered_set` ets tables, which allow to use prefix lookups, hence skipping a supervisor call and replace `count`-ets-lookups with a single one.


sneako (approved):
Good idea!

PR #370, opened 2026-04-22
Retry on not ready pools

jtsmills:
HTTP/2 pools only register in the registry once they have reached the `connected` state. First requests to an on-demand pool can hit `pool_not_available`. The supervisor starts the pool, but TLS handshake may not be completed by the time `get_pool` checks the registry.

**Changes:**
Add a short retry (3 attempts, 100ms apart) for `:not_ready` pools. This matches the existing retry pattern for `:read_only`

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -822,6 +822,10 @@ defmodule Finch do
             other
         end
 
+      :not_ready when retries > 0 ->
+        Process.sleep(100)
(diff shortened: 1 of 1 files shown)


sneako:
Thanks for looking in to this, but this implementation affects all pools (h1 included), and I don't want to introduce this 100ms sleep for all users of Finch. I would rather fail fast and let the called decide if and when they want to retry.

PR #377, opened 2026-06-04
Add certs_keys, certfile, keyfile to dropped TLS options

eErr0Re:
Making an HTTP request from default pool with these options in `transport_opts` results in `:badarg` error from `:gen_tcp`.

Added these keys to the list of TLS options to drop from `transport_opts` when using `:http` scheme.


sneako:
Thank you!

PR #379, opened 2026-06-16
Fix race condition when dynamically starting pool supervisors

Gazler:
Before, the request path did this:

1. Look for a ready pool worker in the pool registry.
2. If none exists, check the supervisor registry.
3. If a supervisor exists but no worker is registered yet, return :not_ready.

There was a race condition, since one process could be starting the supervisor, whilst another would see the supervisor exists before the workers had registered. This would result in a `:pool_not_available`.

To fix this, after `start_pool` returns, we look for the ready pool again.

This uses starting the pool supervisor under the DynamicSupervisor (in start_pool/3) as a synchronization point.

So the path is now:

1. Look for a ready pool worker in the poll registry.
2. If none exists and dynamic start is allowed, call start_pool/3 which calls DynamicSupervisor.start_child/2.
3. If another process already started it, OTP returns {:error, {:already_started, pid}}, which we already treat as :ok.

The test has been updated to assert all of the requests succeed, instead of just asserting pool count.

### DIFF
--- a/lib/finch/pool/manager.ex
+++ b/lib/finch/pool/manager.ex
@@ -101,18 +101,23 @@ defmodule Finch.Pool.Manager do
   defp do_get_pool(registry_name, pool, start_pool?, opts) do
     pool_name = Finch.Pool.to_name(pool)
 
-    case Registry.lookup(registry_name, pool_name) do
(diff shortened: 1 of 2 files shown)


sneako:
Thank you @Gazler !

PR #382, opened 2026-06-19
Add Finch.start_link spec and Finch.pool_opt,pool_opts types

wojtekmach:
In Req I'm adding `Req.request(url, finch: opts)` where opts are _a combination_ of Finch.build, Finch.request, and Finch.start_link options, see https://github.com/wojtekmach/req/pull/542#issuecomment-4742745577.

I'd like to mention pool options by name in the docs because they can't be all mixed and matched (pool options cannot be given when using a named pool).

I _think_ this is the type of single list of options that's then distributed to the three buckets:

```elixir
[{:name, Finch.name()} | Finch.request_opt() | Finch.Request.build_opt()] |
  [Finch.pool_opt() | Finch.request_opt() | Finch.Request.build_opt()]
```

(I'm actually not sure if I'm gonna use above type in my docs because it's pretty complicated and so not sure if that useful.)

### DIFF
--- a/lib/finch.ex
+++ b/lib/finch.ex
@@ -18,6 +18,7 @@ defmodule Finch do
   @pool_config_schema [
     protocols: [
       type: {:list, {:in, [:http1, :http2]}},
+      type_spec: quote(do: [:http1 | :http2]),
       doc: """
       The type of connections to support.
 
@@ -156,6 +157,8 @@ defmodule Finch do
     ]
   ]
 
+  @pool_schema NimbleOptions.new!(@pool_config_schema)
+
   @typedoc """
   The `:name` provided to Finch in `start_link/1`.
   """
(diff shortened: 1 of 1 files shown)


wojtekmach:
As an aside, and I vaguely remember this was maybe brought up already?, here's an idea:

1. Add `%Finch.Request{opts: []}` field
2. Unify `Finch.request_opt` and `Finch.Request.build_opt` into just `Finch.request_opt`

This way, in Finch.request, `Keyword.validate!(opts, allowed)` would no longer necessarily be in the hot path so to speak (cc @sabiwara), if users pass all the options to Finch.build, what is a left is a no-op `Keyword.validate!([], allowed)`.

And yeah, my type above could become:

```elixir
[{:name, Finch.name()} | Finch.request_opt()] |
  [{Finch.pool_opt() | Finch.request_opt()}]
```

That said, it's rather rare for structs to carry options so this needs consideration!

sneako (approved):

PR #387, opened 2026-08-27
Fix `MatchError` when cancelling while HTTP/2 pool is `disconnected`

britto:
We've spotted a few of these uncaught `MatchError`s recently in one of our production systems that uses an HTTP/2 pool. The stack trace led me to line 64 (88 also affected) and I took a stab at it.

The `disconnected` state replied with an error tuple to `{:cancel, ref}` calls, but both callers (`request` and `async_request`) pattern-match strictly on `:ok`, causing a `MatchError`. Now we reply with `:ok`, consistent with the other three states.

Worth noting that raising a `MatchError` at that point prevented `clean_responses` from being called, which could possibly cause pileups in the process mailbox and other sorts of abnormal behavior.


sneako:
Thank you!

PR #392, opened 2026-09-14
Close terminated HTTP/1 connections outside the pool process

ericmj:
`terminate_worker/3` runs in the pool process and calls `Conn.close/1`. For TLS that ends in `ssl:close/1`, which does `shutdown(write)` and then a `recv` with a 5 s default timeout, waiting for the peer's close_notify. If the peer doesn't answer, the pool process sits in that recv for 5 s and every checkout queues behind it. The default `pool_timeout` is also 5 s, so callers raise the "excess queuing" error while the pool has free connections. `conn_max_idle_time` makes this easy to hit because it puts the close on the checkout path.

We hit it on hex.pm: two worker pods with a handful of requests in flight each, both raising at the same second right after Fastly dropped some connections.

The close now runs in a spawned process. It doesn't need to own the socket, and closing an already closed socket is a no-op, so there's nothing for that process to fail on.

dashbitco/nimble_pool#45 traced the same stall to `ssl_gen_statem` in 2024 and closed with "spawn a process for termination in Finch". #353 looks like the same thing.

Tests: a TLS server behind a TCP relay that stops forwarding after the first exchange but keeps the sockets open, so the client's close never gets an answer. After `conn_max_idle_time` passes, two concurrent requests both raise on `main` and both succeed here. Also a unit test that `terminate_worker/3` returns immediately with a transport whose close sleeps, and a `MockSocketServer` test that the spawned close still reaches the peer. `MockSocketServer` called `:inet.port/1` on a TLS listen socket, which fails, so that's fixed too.


sneako (approved):
Great find, thank you @ericmj !

